PR #17, opened 2021-08-10
feat: add operator address that can mint and burn tokens

sammy1991106:
Add a new operator address in the Gateway contract, who is allowed to mint and burn tokens. In reality, the operator will be set to the Ethereum's secondary key's address.


haiyizxx (approved):
looks good

PR #21, opened 2021-09-21
refactor: Cleanup

deluca-mike:
- replace privates with publics to remove function redundancy
- use `Contract.function.selector` instead of inline keccak
- simplify if-reverts into requires
- ERC20 constructor now takes decimals as an argument, making `_setupDecimals` redundant
- make `decimals` immutable (instead of a storage slot)
- more minor gas optimizations

### DIFF
--- a/contracts/src/AxelarGateway.sol
+++ b/contracts/src/AxelarGateway.sol
@@ -11,22 +11,24 @@ contract AxelarGateway {
         address indexed previousOwner,
         address indexed newOwner
     );
+
     event OperatorshipTransferred(
         address indexed previousOperator,
         address indexed newOperator
     );
+
     event TokenDeployed(string symbol, address tokenAddress);
 
-    address private _prevOwner;
-    address private _owner;
-    address private _operator;
-    address private _prevOperator;
+    address public prevOwner;
+    address public owner;
+    address public operator;
+    address public prevOperator;
 
     mapping(string => bytes4) private _commandSelectors;
     mapping(string => address) private _commandAddresses;
     mapping(bytes32 => bool) private _commandExecuted;
 
-    mapping(string => address) private _tokenAddresses;
+    mapping(string => address) public tokenAddresses;
 
     modifier onlySelf() {
         require(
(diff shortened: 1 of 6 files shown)


contracts/src/ECDSA.sol
```diff
@@ -26,12 +26,10 @@ library ECDSA {
     function recover(bytes32 hash, bytes memory signature)
         internal
         pure
-        returns (address)
+        returns (address signer)
```
sammy1991106, line 29: I'm not sure I like this syntax. Golang does it too and I always try to avoid it.

contracts/src/Ownable.sol
```diff
@@ -32,7 +25,6 @@ abstract contract Ownable {
             'Ownable: new owner is the zero address'
         );
 
-        emit OwnershipTransferred(_owner, newOwner);
-        _owner = newOwner;
+        emit OwnershipTransferred(owner, owner = newOwner);
```
sammy1991106, line 28: Not a fan of this syntax either. I find it confusing.

contracts/src/Ownable.sol
```diff
@@ -3,26 +3,19 @@
...
     constructor() {
-        address msgSender = msg.sender;
-        _owner = msgSender;
-
-        emit OwnershipTransferred(address(0), msgSender);
-    }
-
-    function owner() public view returns (address) {
-        return _owner;
+        emit OwnershipTransferred(address(0), owner = msg.sender);
```
sammy1991106, line 14: Here too

sammy1991106, line 14: I'd rather do it with 2 lines. Does this even save any gas?

contracts/src/ERC20.sol
```diff
@@ -30,75 +30,27 @@ import './IERC20.sol';
  * allowances. See {IERC20-approve}.
  */
 contract ERC20 is Context, IERC20 {
-    mapping(address => uint256) private _balances;
+    mapping(address => uint256) public override balanceOf;
```
sammy1991106, line 33: Here's a lot of change to openzepplin's code which I've wanted to avoid 😢

contracts/src/AxelarGateway.sol
```diff
@@ -194,17 +161,15 @@ contract AxelarGateway {
                 decimals,
                 cap
             );
-        address tokenAddress = address(token);
-        _tokenAddresses[symbol] = tokenAddress;
 
-        emit TokenDeployed(symbol, tokenAddress);
+        emit TokenDeployed(symbol, tokenAddresses[symbol] = address(token));
```
sammy1991106, line 165: same here

contracts/src/AxelarGateway.sol
```diff
@@ -148,19 +115,18 @@ contract AxelarGateway {
...
                 );
 
-            if (!success) {
-                revert(
-                    string(
-                        abi.encodePacked(
-                            'AxelarGateway: command failed with error: ',
-                            result
-                        )
+            require(
```
sammy1991106, line 121: 👍

sammy1991106 (approved):

PR #28, opened 2021-11-01
feat: multisig axelar gateway

sammy1991106:


cgorenflo (approved):
Please wait for someone else to approve as well though ;)

deluca-mike (changes requested):
- threshold is sometimes a `uint256`, sometimes a `uint8`, which matters when `abi.encodePacked`. Double-check.
- Proxy slots should use EIP-1967 in order to be compatible with explorers such as etherscan.
- `Context` not really needed.
- The `AxelarGatewayProxy` constructor should not deploy the implementation. Rather, the implementation should be deployed and the Proxy directed to it, in a separate transaction (the constructor).

PR #29, opened 2021-11-13
refactor: audit, cleanup, fixes, and readability

deluca-mike:


contracts/src/AxelarGateway.sol
```diff
@@ -168,67 +161,106 @@ abstract contract AxelarGateway is IAxelarGateway, EternalStorage {
...
+        //        if `_update` is entered within the context of _this_ implementation itself. Consider directly calling `forceUpdate`.
+        (bool success, ) =
+            newImplementation.delegatecall(abi.encodeWithSelector(IAxelarGateway.setup.selector, setupParams));
         require(success, 'SETUP_FAILED');
 
-        setAddress(KEY_IMPLEMENTATION, newVersion);
+        _setImplementation(newImplementation);
 
-        emit Updated(address(this), newVersion);
+        emit Updated(address(this), newImplementation);
```
sammy1991106, line 198: Maybe let's pick up the `Upgraded` event defined in https://eips.ethereum.org/EIPS/eip-1967? `address(this)` here would be the proxy itself so it's not correct anyway.

contracts/src/AxelarGatewayMultisig.sol
```diff
@@ -2,312 +2,473 @@
...
+        _setUint(KEY_OWNER_EPOCH, ownerEpoch);
+    }
 
-        _deployToken(name, symbol, decimals, cap);
+    function _setOwner(
+        uint256 ownerEpoch,
+        uint256 index,
+        address account
+    ) internal {
+        // AUDIT: Should have `require(account != address(0), 'ZERO_ADDR');` like Singlesig?
```
sammy1991106, line 138: sure I agree

contracts/src/AxelarGatewayMultisig.sol
```diff
@@ -2,312 +2,473 @@
...
+        _setOwnerEpoch(++ownerEpoch);
+        _setOwners(ownerEpoch, newOwners, newThreshold);
     }
 
-    function _getOperatorsIndex() internal view returns (uint256) {
-        return getUint(KEY_OPERATORS_INDEX);
+    function transferOperatorship(address[] memory signers, bytes memory params) external onlySelf {
+        (address[] memory newOperators, uint256 newThreshold) = abi.decode(params, (address[], uint256));
+
+        // NOTE: Should this be `_areValidOwnersInEpoch` or `_areValidOperatorsInEpoch`?
```
sammy1991106, line 371: It is correct here to use `_areValidOwnersInEpoch`.

contracts/src/AxelarGateway.sol
deluca-mike: Only if you're ok with `proposeUpdate` being changed to `proposeUpgrade`. But I do agree on the standardized `Upgraded` event.

sammy1991106: yes I'll do that.

sammy1991106 (approved):

PR #51, opened 2022-01-25
feature(AxelarGateway): external contracts method calling

re1ro:
* [x] New `callContract` and `callContractWithToken` gateway methods for cross-chain communication
* [x] Adding `approveContractCallWithMint` and `approveContractCall` to approve commands on a destination chain
* [x] New `validateContractCall` and `validateContractCallAndMint` gateway methods for validating contract calls  on a destination chain
* [x] Unit tests

### DIFF
--- a/src/interfaces/IAxelarGateway.sol
+++ b/src/interfaces/IAxelarGateway.sol
@@ -18,9 +18,9 @@ interface IAxelarGateway {
 
     event TokenDeployed(string symbol, address tokenAddresses);
 
-    event ContractCallApproved(bytes32 indexed commandId, string sourceChain, string sourceAddress, address indexed contractAddress, bytes32 payloadHash);
+    event ContractCallApproved(bytes32 indexed commandId, string sourceChain, string sourceAddress, address indexed contractAddress, bytes32 indexed payloadHash);
 
-    event ContractCallApprovedWithMint(bytes32 indexed commandId, string sourceChain, string sourceAddress, address indexed contractAddress, bytes32 payloadHash, string indexed symbol, uint256 amount);
+    event ContractCallApprovedWithMint(bytes32 indexed commandId, string sourceChain, string sourceAddress, address indexed contractAddress, bytes32 indexed payloadHash, string symbol, uint256 amount);
 
     event TokenFrozen(string indexed symbol);
 


src/AxelarGateway.sol
```diff
@@ -161,6 +168,29 @@ abstract contract AxelarGateway is IAxelarGateway, AdminMultisigBase {
...
+    }
+
+    function _approveContractCall(
+        address contractAddress,
+        bytes32 payloadHash
+    ) internal {
+        // using keyed hashing to prevent hash forging for minted/direct call approvals
+        _setContractCallApproved(contractAddress, keccak256(abi.encode(ContractCallHashKey.WithoutToken, payloadHash)));
+
+        emit ContractCallApproved(contractAddress, ContractCallHashKey.WithToken, payloadHash);
```
sammy1991106, line 191:
```suggestion
        emit ContractCallApproved(contractAddress, ContractCallHashKey.WithoutToken, payloadHash);
```

src/AxelarGateway.sol
```diff
@@ -161,6 +168,29 @@ abstract contract AxelarGateway is IAxelarGateway, AdminMultisigBase {
         }
     }
 
+    function _mintTokenAndApproveContractCall(
```
sammy1991106, line 171: Can we merge `_approveContractCall` and `_mintTokenAndApproveContractCall` into one function to save some gas on deploying the gateway? Also, I don't think we need to events like `MintedAndContractCallApproved` and `ContractCallApproved`. One can already figure out that some token was minted in the same tx as `_mintToken` emits a `Transfer` event. Just `event ContractCallApproved(address contractAddress, ContractCallHashKey hashKey, bytes32 payloadHash);` should be sufficient.

sammy1991106 (changes requested):

src/AxelarGateway.sol
re1ro: I think it's a good idea. I was originally trying to be explicit about what is happening, especially if we have many mint commands in one `_execute` transaction. But you are right, it is possible to determine previous mint event from the tx event list

src/AxelarGateway.sol
```diff
@@ -178,6 +186,34 @@ abstract contract AxelarGateway is IAxelarGateway, AdminMultisigBase {
...
+        uint256 amount,
+        address contractAddress,
+        bytes32 payloadHash
+    ) internal {
+        if (amount > 0) {
+            _mintToken(symbol, contractAddress, amount);
+            _setContractCallApproved(
+                contractAddress,
+                keccak256(
+                    abi.encode(ContractCallHashKey.WithToken, tokenAddresses(symbol), amount, payloadHash)
```
deluca-mike, line 200: Why is a `ContractCallHashKey` enum even needed if the resulting hashes are already different given that "with token: has 2 additional parameters?

src/AxelarGatewayMultisig.sol
```diff
@@ -346,6 +346,21 @@ contract AxelarGatewayMultisig is IAxelarGatewayMultisig, AxelarGateway {
...
+            (string, address, uint256, bytes32)
+        );
+
+        _mintTokenAndApproveContractCall(symbol, amount, contractAddress, payloadHash);
+    }
+
+    function approveContractCall(bytes calldata params) external onlySelf {
+        (address contractAddress, bytes32 payloadHash) = abi.decode(params, (address, bytes32));
+
+        _mintTokenAndApproveContractCall('', 0, contractAddress, payloadHash);
```
deluca-mike, line 361: This kind of "overloading" of functions is a programming anti-pattern that should be avoided. No gas is being saved and an `if` has to happen soon or later. Better to just have 2 internal functions instead of one that does 2 different things depending on how it is called, and then having one caller pass in dummy data or zeros for some arguments.

deluca-mike (changes requested):

src/test/ExternalExecutor.sol
```diff
@@ -0,0 +1,35 @@
+// SPDX-License-Identifier: MIT
```
sammy1991106, line 1: Later on, these contracts used for testing should be placed in `./src/test/contracts`

re1ro: It is not trivial for our waffle configuration, we would need a separate one for test contacts. And for each previous EVM version we are testing agains as well

PR #55, opened 2022-01-28
feat: deposit handler

sammy1991106:


re1ro (approved):
Looks great! I like that DepositHandler can do just about anything. And would support NFT out of the box as well!

sammy1991106:
> Looks great! I like that DepositHandler can do just about anything. And would support NFT out of the box as well!

@re1ro The original burner came with an idea that it's a commitment to the users that once they've sent their tokens to the burner address, the only thing that can ever happen to them is being burnt. Therefore, users wouldn't have to worry about us abusing our power to "steal" their supposedly-burnt tokens. Now, we are basically lifting all those restrictions that were intentionally put in place. Just wanna make sure we all understand the tradeoffs.

deluca-mike (approved):

re1ro:
@sammy1991106 I understand that. Since the handler can be called only from the gateway, the gateway becomes the point of trust here

PR #58, opened 2022-02-08
feature(Gateway): sendToken directly from the Gateway

re1ro:
* [x] `BurnableMintableCappedERC20`: adding new `burnFrom` method requiring users approval
* [x] `AxelarGateway`: adding new `sendToken` method that burns token and emits `TokenSent` event
* [x] Unit tests


src/AxelarGateway.sol
```diff
@@ -43,6 +49,45 @@ abstract contract AxelarGateway is IAxelarGateway, AdminMultisigBase {
...
 
+    /******************\
+    |* Public Methods *|
+    \******************/
+
+    function sendToken(
+        uint256 destinationChain,
+        string memory symbol,
+        uint256 amount
+    ) external returns (bool) {
```
sammy1991106, line 60:
```suggestion
    ) external {
```
The returning-bool practice is pretty useless and is just a legacy from erc20.

src/AxelarGateway.sol
```diff
@@ -128,11 +173,12 @@ abstract contract AxelarGateway is IAxelarGateway, AdminMultisigBase {
             // If token address is no specified, it indicates a request to deploy one.
             bytes32 salt = keccak256(abi.encodePacked(symbol));
             tokenAddress = address(new BurnableMintableCappedERC20{ salt: salt }(name, symbol, decimals, cap));
+            _setTokenType(symbol, TokenType.InternalBurnableFrom);
         } else {
             // If token address is specified, ensure that there is a contact at the specified addressed.
-            require(tokenAddress.code.length != uint256(0), 'NOT_TOKEN');
+            require(tokenAddress.codehash != keccak256(''), 'NOT_TOKEN');
```
sammy1991106, line 179: Create a constant for `keccak256('')` so that it doesn't have to hash it again every time.

src/AxelarGateway.sol
```diff
@@ -43,6 +49,45 @@ abstract contract AxelarGateway is IAxelarGateway, AdminMultisigBase {
...
+        TokenType tokenType = _getTokenType(symbol);
+        bytes memory tokenCallData;
+        bytes32 salt;
+
+        if (tokenType == TokenType.External) {
+            tokenCallData = abi.encodeWithSelector(IERC20.transferFrom.selector, msg.sender, address(this), amount);
+        } else if (tokenType == TokenType.InternalBurnableFrom) {
+            tokenCallData = abi.encodeWithSelector(BurnableMintableCappedERC20.burnFrom.selector, msg.sender, amount);
+        } else if (tokenType == TokenType.InternalBurnable) {
+            salt = keccak256(abi.encodePacked(destinationChain, msg.sender, tokenAddress));
```
sammy1991106, line 73: At this point, by bother creating a different deposit address for each user? Why not just have everyone to share the same as 
```
address depositAddress = BurnableMintableCappedERC20(tokenAddress).depositAddress(bytes32(0));
```

sammy1991106 (changes requested):

src/AxelarGateway.sol
re1ro: I think you are right, and this will always work. But still feels dangerous 😄

src/AxelarGateway.sol
```diff
@@ -43,6 +49,45 @@ abstract contract AxelarGateway is IAxelarGateway, AdminMultisigBase {
...
+        }
+
+        (bool success, bytes memory returnData) = tokenAddress.call(tokenCallData);
+        require(success && (returnData.length == uint256(0) || abi.decode(returnData, (bool))), 'BURN_FAIL');
+
+        if (tokenType == TokenType.InternalBurnable) {
+            BurnableMintableCappedERC20(tokenAddress).burn(salt);
+        }
+
+        emit TokenSent(destinationChain, symbol, amount);
```
sammy1991106, line 86: If the contract cannot validate `destinationChain`, what does axelar network do if it doesn't view it as one of those being supported due to whatever reason, e.g. misconfiguration or user error?

src/AxelarGateway.sol
```diff
@@ -43,6 +49,45 @@ abstract contract AxelarGateway is IAxelarGateway, AdminMultisigBase {
         _;
     }
 
+    /******************\
+    |* Public Methods *|
+    \******************/
+
+    function sendToken(
+        uint256 destinationChain,
```
sammy1991106, line 57:
```suggestion
        uint256 destinationChainId,
```

sammy1991106 (changes requested):

src/AxelarGateway.sol
re1ro: That's a very good point. I think we need to add dynamic adding and removing allowed destination chains

re1ro: Added support for that

src/AxelarGateway.sol
```diff
@@ -128,11 +180,12 @@ abstract contract AxelarGateway is IAxelarGateway, AdminMultisigBase {
             // If token address is no specified, it indicates a request to deploy one.
             bytes32 salt = keccak256(abi.encodePacked(symbol));
             tokenAddress = address(new BurnableMintableCappedERC20{ salt: salt }(name, symbol, decimals, cap));
+            _setTokenType(symbol, TokenType.InternalBurnableFrom);
         } else {
             // If token address is specified, ensure that there is a contact at the specified addressed.
-            require(tokenAddress.code.length != uint256(0), 'NOT_TOKEN');
+            require(tokenAddress.codehash != EMPTY_HASH, 'NOT_TOKEN');
```
deluca-mike, line 186: Cheaper to check `tokenAddress.code.length != 0`.

src/BurnableMintableCappedERC20.sol
```diff
@@ -45,6 +45,11 @@ contract BurnableMintableCappedERC20 is MintableCappedERC20 {
         _burn(account, balanceOf[account]);
     }
 
+    function burnFrom(address account, uint256 amount) public onlyOwner {
+        _approve(account, owner, allowance[account][owner] - amount);
+        _burn(account, amount);
+    }
+
```
deluca-mike, line 52: This now requires approval. Perhaps now is a good tie to add [ERC20 Permits](https://docs.openzeppelin.com/contracts/4.x/api/token/erc20#ERC20Permit) to the contracts so that this can be done in one tx.

deluca-mike (changes requested):

src/AxelarGateway.sol
```diff
@@ -16,6 +16,12 @@ abstract contract AxelarGateway is IAxelarGateway, AdminMultisigBase {
         Operator
     }
 
+    enum TokenType {
+        InternalBurnable,
+        InternalBurnableFrom,
+        External
+    }
```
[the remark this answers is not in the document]
re1ro: We don't have ERC20 support deployed yet. So there are no `1` values in storage yet

src/AxelarGateway.sol
re1ro: You are tight, reading just bytesize was implemented in `0.8.0`

src/AxelarGateway.sol
```diff
@@ -43,6 +52,41 @@ abstract contract AxelarGateway is IAxelarGateway, AdminMultisigBase {
         _;
     }
 
+    /******************\
+    |* Public Methods *|
+    \******************/
+
+    function sendToken(
```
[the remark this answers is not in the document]
re1ro: Implemented

src/BurnableMintableCappedERC20.sol
re1ro: Added permits in this PR

src/BurnableMintableCappedERC20.sol
```diff
@@ -45,6 +61,36 @@ contract BurnableMintableCappedERC20 is MintableCappedERC20 {
...
+    function permit(
+        address owner,
```
sammy1991106, line 70: Choose a different variable name than `owner` cuz it's already used as a storage variable and looks confusing.

waffle.json
```diff
@@ -7,7 +7,7 @@
     "evmVersion": "london",
     "optimizer": {
       "enabled": true,
-      "runs": 200
+      "runs": 10
```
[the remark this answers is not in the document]
re1ro: Because we already hit the contract size limit
```
Warning: Contract code size exceeds 24576 bytes (a limit introduced in Spurious Dragon). 
This contract may not be deployable on mainnet. 
Consider enabling the optimizer (with a low "runs" value!), 
turning off revert strings, or using libraries.
```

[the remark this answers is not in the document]
sammy1991106: Actually, I'd love to turn off revert strings if that's possible. That sounds lovely.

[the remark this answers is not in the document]
re1ro: After removing chainId whitelisting I was able to revert runs back to `200`

test/AxelarGatewayMultisig.js
```diff
@@ -893,5 +893,59 @@ describe('AxelarGatewayMultisig', () => {
           });
       });
     });
+
+    describe('send token from gateway', () => {
+      it('should burn token and emit an event', async () => {
```
sammy1991106, line 898:
```suggestion
    describe('sendToken, () => {
      it('should burn internal token and emit an event', async () => {
```
so that you test all kinds of things about sendToken in the describe

sammy1991106: need to also test external token here

test/AxelarGatewaySinglesig.js
```diff
@@ -1311,5 +1310,59 @@ describe('AxelarGatewaySingleSig', () => {
           });
       });
     });
+
+    describe('send token from gateway', () => {
```
sammy1991106, line 1314: same here

test/BurnableMintableCappedERC20.js
```diff
@@ -0,0 +1,160 @@
...
+      await expect(await token.connect(userWallet).approve(spender, amount))
+        .to.emit(token, 'Approval')
+        .withArgs(issuer, spender, amount);
+
+      await expect(await gateway.connect(userWallet).sendToken(2, ownerWallet.address, tokenSymbol, amount))
+        .to.emit(token, 'Transfer')
+        .withArgs(issuer, ADDRESS_ZERO, amount);
+    });
+  });
+  describe('EIP-2612: approve permit', () => {
```
sammy1991106, line 110:
```suggestion

  describe('permit', () => {
```

test/BurnableMintableCappedERC20.js
```diff
@@ -0,0 +1,160 @@
...
+
+    const tokenAddress = await gateway.tokenAddresses(tokenSymbol)
+    token = new Contract(
+      tokenAddress,
+      BurnableMintableCappedERC20.abi,
+      wallets[0],
+    );
+  });
+
+  describe('burning from account with given approve', () => {
```
sammy1991106, line 95:
```suggestion
  describe('burnFrom, () => {
```

test/BurnableMintableCappedERC20.js
```diff
@@ -0,0 +1,160 @@
...
+  const operators = wallets.slice(3, 6);
+  const admins = wallets.slice(6, 9);
+  const threshold = 2;
+
+  const ownerWallet = owners[0]
+  const userWallet = admins[0]
+
+  const tokenName = 'Test Token';
+  const tokenSymbol = 'TEST';
+  let gateway;
```
sammy1991106, line 35: Do you absolutely need a gateway to test `BurnableMintableCappedERC20.js?

re1ro: Moved the `burnFrom` back into `MintableCappedERC20` so we don't need the gateway to deploy it

test/AxelarGatewayMultisig.js
```diff
@@ -893,5 +894,117 @@ describe('AxelarGatewayMultisig', () => {
           });
       });
     });
+
+    describe('send token from gateway', () => {
+      it('should burn internal token and emit an event', async () => {
```
deluca-mike, line 899: We should test both types of internal token.

test/AxelarGatewaySinglesig.js
deluca-mike: We should test both types of internal tokens.

deluca-mike (approved):
Just lacks one internal token test for each contract.

test/AxelarGatewayMultisig.js
re1ro: We will make it a part of integration tests in the future

sammy1991106 (approved):

PR #61, opened 2022-02-16
Adding support for tokenType, burnFrom and permit

re1ro:
* [x] build: dev and prod builds with `"revertStrings": "strip"`
* [x] BurnableMintableCappedERC20: burnFrom + permit
* [x] AxelarGateway: using tokenType + _callERC20Token util
* [x] AxelarGateway refactor: using block.chainid


src/BurnableMintableCappedERC20.sol
```diff
@@ -18,7 +27,15 @@ contract BurnableMintableCappedERC20 is MintableCappedERC20 {
         string memory symbol,
         uint8 decimals,
         uint256 capacity
-    ) MintableCappedERC20(name, symbol, decimals, capacity) {}
+    ) MintableCappedERC20(name, symbol, decimals, capacity) {
+        DOMAIN_SEPARATOR = keccak256(
+            abi.encode(
+                keccak256('EIP712Domain(uint256 chainId,address verifyingContract)'),
```
sammy1991106, line 33: Why is this not a constant?

src/BurnableMintableCappedERC20.sol
```diff
@@ -45,6 +62,36 @@ contract BurnableMintableCappedERC20 is MintableCappedERC20 {
         _burn(account, balanceOf[account]);
     }
 
+    function burnFrom(address account, uint256 amount) public onlyOwner {
```
sammy1991106, line 65: I don't see testing for `burnFrom` and `permit`?

sammy1991106 (changes requested):

src/BurnableMintableCappedERC20.sol
re1ro: Good idea! I will move that hash in constants

src/BurnableMintableCappedERC20.sol
re1ro: I should push them today

src/BurnableMintableCappedERC20.sol
```diff
@@ -45,6 +70,36 @@ contract BurnableMintableCappedERC20 is MintableCappedERC20 {
...
+
+        bytes32 digest = keccak256(
+            abi.encodePacked(
+                EIP191_PREFIX_FOR_EIP712_STRUCTURED_DATA,
+                DOMAIN_SEPARATOR,
+                keccak256(abi.encode(PERMIT_SIGNATURE_HASH, issuer, spender, value, nonces[issuer]++, deadline))
+            )
+        );
+
+        address recoveredAddress = ecrecover(digest, v, r, s);
```
sammy1991106, line 97: The ECDSA lib implemented by OZ requires s to be in the lower range and requires v to be 27 or 28. Should we enforce the same? @deluca-mike thought? https://github.com/axelarnetwork/solidity-cgp-gateway/blob/ce22184bcfe2481c60cce20d841f435947a4d5d3/src/ECDSA.sol#L44

src/BurnableMintableCappedERC20.sol
```diff
@@ -45,6 +70,36 @@ contract BurnableMintableCappedERC20 is MintableCappedERC20 {
...
+        require(recoveredAddress == issuer, 'INVALID_SIGNATURE');
```
sammy1991106, line 98:
```suggestion
        require(recoveredAddress != address(0), 'INV_SIG');
        require(recoveredAddress == issuer, 'INV_SIG');
```
solidity's ecrecover returns `address(0)` instead of revert on error https://docs.soliditylang.org/en/v0.8.9/units-and-global-variables.html?highlight=ecrecover#mathematical-and-cryptographic-functions

sammy1991106 (approved):

src/BurnableMintableCappedERC20.sol
re1ro: @sammy1991106 I am adding those sanity checks as the best security practice

PR #77, opened 2022-03-28
fix(Multisig): transferOperatorship using operatorEpoch

re1ro:
* [x] `OperatorshipTransferred` event: using operatorEpoch for getting old threshold  value
* [x] Tests: coverage for `OwnershipTransferred` and `OperatorshipTransferred` events

### DIFF
--- a/test/AxelarGatewayMultisig.js
+++ b/test/AxelarGatewayMultisig.js
@@ -600,7 +600,7 @@ describe('AxelarGatewayMultisig', () => {
 
     describe('command transferOwnership', () => {
       it('should owners to transfer ownership', async () => {
-        const makeTransferCommand = newThreshold =>
+        const makeTransferCommand = (newThreshold) =>
           arrayify(
             defaultAbiCoder.encode(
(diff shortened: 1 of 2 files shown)


test/AxelarGatewayMultisig.js
```diff
@@ -599,40 +599,58 @@ describe('AxelarGatewayMultisig', () => {
...
-        const data = arrayify(
-          defaultAbiCoder.encode(
-            ['uint256', 'uint256', 'bytes32[]', 'string[]', 'bytes[]'],
-            [
-              CHAIN_ID,
-              ROLE_OWNER,
-              [getRandomID()],
-              ['transferOwnership'],
+      it('should owners to transfer ownership', async () => {
+        const makeTransferCommand = (newThreshold) =>
```
sammy1991106, line 603: I would extract this function to the top of the file so that you don't need to define it twice inline

sammy1991106 (approved):

test/AxelarGatewayMultisig.js
re1ro: Will do that to make things cleaner

PR #80, opened 2022-04-05
feature(Gateway): freezing external tokens

re1ro:
* [x] `mintToken`, `burnToken`: checking for external token freeze
* [x] `_burnTokenFrom`: checking for external token freeze

### DIFF
--- a/test/AxelarGatewaySinglesig.js
+++ b/test/AxelarGatewaySinglesig.js
@@ -1143,100 +1143,6 @@ describe('AxelarGatewaySingleSig', () => {
     });
 
     describe('batch commands', () => {
-      it('should support external ERC20 token', () => {
-        const name = 'test';
-        const symbol = 'test';
-        const decimals = 16;
-        const capacity = 0;
-
-        return deployContract(ownerWallet, MintableCappedERC20, [
-          name,
-          symbol,
-          decimals,
-          capacity,
-        ]).then(async (token) => {
-          const amount = 10000;
-          await token.mint(nonOwnerWallet.address, amount);
-
-          const deployTokenData = arrayify(
-            defaultAbiCoder.encode(
-              ['uint256', 'uint256', 'bytes32[]', 'string[]', 'bytes[]'],
-              [
-                CHAIN_ID,
-                ROLE_OWNER,
-                [getRandomID()],
-                ['deployToken'],
-                [
-                  defaultAbiCoder.encode(
-                    ['string', 'string', 'uint8', 'uint256', 'address'],
-                    [name, symbol, decimals, capacity, token.address],
-                  ),
-                ],
-              ],
-            ),
-          );
-          await getSignedExecuteInput(deployTokenData, ownerWallet).then(
-            (input) =>
-              expect(contract.execute(input))
-                .to.emit(contract, 'TokenDeployed')
-                .withArgs(symbol, token.address),
-          );
-
-          const salt = randomBytes(32);
-          const depositHandlerAddress = getCreate2Address(
-            contract.address,
-            salt,
-            keccak256(`0x${DepositHandler.bytecode}`),
-          );
-          await token
-            .connect(nonOwnerWallet)
-            .transfer(depositHandlerAddress, amount);
-
-          const burnTokenData = arrayify(
-            defaultAbiCoder.encode(
-              ['uint256', 'uint256', 'bytes32[]', 'string[]', 'bytes[]'],
-              [
-                CHAIN_ID,
-                ROLE_OWNER,
-                [getRandomID()],
-                ['burnToken'],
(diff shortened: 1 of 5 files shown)


Foivos (approved):

test/AxelarGatewaySinglesig.js
```diff
@@ -1320,6 +1226,252 @@ describe('AxelarGatewaySingleSig', () => {
...
+        );
+
+        await getSignedExecuteInput(getMintTokenData(), ownerWallet).then(
+          (input) =>
+            expect(contract.execute(input)).not.to.emit(token, 'Transfer'),
+        );
+
+        for (const adminWallet of adminWallets.slice(0, 3)) {
+          await contract.connect(adminWallet).unfreezeToken(symbol);
+        }
```
sammy1991106, line 1424:
```suggestion
        await Promise.all(
          adminWallets
            .slice(0, 3)
            .map((admin) => contract.connect(admin).unfreezeToken(symbol)),
        );
```
Doing `for` loop in js is almost never a good idea. It's either `.forEach` if sync or `Promise.all` if async. `Promise.all` is also much more performant than `for` loop if async.

test/AxelarGatewaySinglesig.js
```diff
@@ -1320,6 +1226,252 @@ describe('AxelarGatewaySingleSig', () => {
...
+        );
+
+        await getSignedExecuteInput(getBurnTokenData(), ownerWallet).then(
+          (input) =>
+            expect(contract.execute(input))
+              .to.emit(token, 'Transfer')
+              .withArgs(depositHandlerAddress, contract.address, amount),
+        );
+
+        for (const adminWallet of adminWallets.slice(0, 3)) {
```
sammy1991106, line 1437: same here.

test/AxelarGatewaySinglesig.js
```diff
@@ -1320,6 +1226,252 @@ describe('AxelarGatewaySingleSig', () => {
...
+        );
+
+        await getSignedExecuteInput(getBurnTokenData(), ownerWallet).then(
+          (input) =>
+            expect(contract.execute(input))
+              .not.to.emit(token, 'Transfer')
+              .withArgs(depositHandlerAddress, contract.address, amount),
+        );
+
+        for (const adminWallet of adminWallets.slice(0, 3)) {
```
sammy1991106, line 1453: same here.

test/AxelarGatewaySinglesig.js
```diff
@@ -1426,6 +1578,79 @@ describe('AxelarGatewaySingleSig', () => {
...
+
+      for (const adminWallet of adminWallets.slice(0, 3)) {
+        await contract.connect(adminWallet).freezeToken(tokenSymbol);
+      }
+
+      await expect(
+        contract.sendToken('polygon', destination, tokenSymbol, amount),
+      ).to.be.reverted;
+
+      for (const adminWallet of adminWallets.slice(0, 3)) {
```
sammy1991106, line 1633: same here.

test/AxelarGatewaySinglesig.js
```diff
@@ -1426,6 +1578,79 @@ describe('AxelarGatewaySingleSig', () => {
...
+        await contract.connect(adminWallet).unfreezeToken(tokenSymbol);
+        await contract.connect(adminWallet).freezeAllTokens();
+      }
+
+      await expect(
+        contract.sendToken('polygon', destination, tokenSymbol, amount),
+      ).to.be.reverted;
+
+      for (const adminWallet of adminWallets.slice(0, 3)) {
```
sammy1991106, line 1642: same here.

test/AxelarGatewaySinglesig.js
```diff
@@ -1320,6 +1226,252 @@ describe('AxelarGatewaySingleSig', () => {
...
+                  defaultAbiCoder.encode(
+                    ['string', 'address', 'uint256'],
+                    [symbol, ownerWallet.address, amount],
+                  ),
+                ],
+              ],
+            ),
+          );
+
+        for (const adminWallet of adminWallets.slice(0, 3)) {
```
sammy1991106, line 1406: same here.

sammy1991106 (approved):

test/AxelarGatewaySinglesig.js
re1ro: @sammy1991106  the problem with `Promise.all` or `.forEach` is that it will execute all 3 calls immediately and this might break the TX nonce. `for` loop with async with sequential performance is here on purpose, so we can wait for each call to finish.

sammy1991106: No it won't break nonce. Each admin is a different address.

sammy1991106 (approved):

PR #81, opened 2022-04-05
feature(Gateway): getting admins, owners, operators and their thresholds

re1ro:
### New methods
* [x] `getAdminEpoch()`, `getAdminThreshold(uint256 adminEpoch)`, `admins(uint256 adminEpoch)`
* [x] `getOwnerEpoch()`, `getOwnerThreshold(uint256 ownerEpoch)`
* [x] `getOperatorEpoch()`, `getOperatorThreshold(uint256 operatorEpoch)` 

### Changes methods
* [x] `owners()` -> `owners(uint256 ownerEpoch)`
* [x] `operators()` -> `operators(uint256 operatorEpoch)`

### DIFF
--- a/src/AxelarGateway.sol
+++ b/src/AxelarGateway.sol
@@ -206,6 +206,26 @@ abstract contract AxelarGateway is IAxelarGateway, AdminMultisigBase {
         return getBool(_getIsCommandExecutedKey(commandId));
     }
 
+    /// @dev Returns the current `adminEpoch`.
+    function getAdminEpoch() external view override returns (uint256) {
+        return _adminEpoch();
+    }
+
+    /// @dev Returns the admin threshold for a given `adminEpoch`.
+    function getAdminThreshold(uint256 adminEpoch) external view override returns (uint256) {
+        return _getAdminThreshold(adminEpoch);
+    }
+
+    /// @dev Returns the array of admins within a given `adminEpoch`.
+    function admins(uint256 adminEpoch) external view override returns (address[] memory results) {
+        uint256 adminCount = _getAdminCount(adminEpoch);
+        results = new address[](adminCount);
+
+        for (uint256 i; i < adminCount; i++) {
+            results[i] = _getAdmin(adminEpoch, i);
+        }
+    }
+
     /*******************\
     |* Admin Functions *|
     \*******************/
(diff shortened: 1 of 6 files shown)


src/AxelarGateway.sol
```diff
@@ -206,6 +206,26 @@ abstract contract AxelarGateway is IAxelarGateway, AdminMultisigBase {
         return getBool(_getIsCommandExecutedKey(commandId));
     }
 
+    /// @dev Returns the current `adminEpoch`.
+    function getAdminEpoch() external view override returns (uint256) {
```
sammy1991106, line 210: Why expose this? This is implementation detail.

src/AxelarGatewayMultisig.sol
```diff
@@ -122,9 +122,18 @@ contract AxelarGatewayMultisig is IAxelarGatewayMultisig, AxelarGateway {
         return false;
     }
 
-    /// @dev Returns the array of owners within the current `ownerEpoch`.
-    function owners() public view override returns (address[] memory results) {
-        uint256 ownerEpoch = _ownerEpoch();
+    /// @dev Returns the current `ownerEpoch`.
+    function getOwnerEpoch() external view override returns (uint256) {
```
sammy1991106, line 126: Same here. We should not expose this.

src/AxelarGateway.sol
```diff
@@ -206,6 +206,26 @@ abstract contract AxelarGateway is IAxelarGateway, AdminMultisigBase {
...
+        return _adminEpoch();
+    }
+
+    /// @dev Returns the admin threshold for a given `adminEpoch`.
+    function getAdminThreshold(uint256 adminEpoch) external view override returns (uint256) {
```
sammy1991106, line 215: Do we really care about the threshold of a past epoch ever? If not, I'd rather these only being capable returning the latest.

sammy1991106:
I don't see why we would ever want admin/owner/operator information from the past and did you confirm with @milapsheth that he actually needs them to go that far?

sammy1991106 (changes requested):

src/AxelarGateway.sol
sammy1991106, line 215:
```suggestion
    function adminThreshold(uint256 adminEpoch) external view override returns (uint256) {
```
Since the other getter is call `admins()` instead of `getAdmins()`, also better considering solidity's practice.

sammy1991106 (approved):

sammy1991106 (approved):

PR #86, opened 2022-04-13
Gateway: token deployer storage + upgrade

re1ro:
* [x] Removed `immutable TOKEN_DEPLOYER_IMPLEMENTATION`
* [x] Added `KEY_TOKEN_DEPLOYER_IMPLEMENTATION` storage slot
* [x] Changed proxy constructor and added `upgradeTokenDeployer` method
* [ ] Problem: `AxelarGatewayMultisig` contract code size exceeds 24576 bytes

### DIFF
--- a/src/AxelarGatewayProxy.sol
+++ b/src/AxelarGatewayProxy.sol
@@ -12,9 +12,17 @@ contract AxelarGatewayProxy is EternalStorage {
     /// @dev Storage slot with the address of the current factory. `keccak256('eip1967.proxy.implementation') - 1`.
     bytes32 internal constant KEY_IMPLEMENTATION =
         bytes32(0x360894a13ba1a3210667c828492db98dca3e2076cc3735a920a3ca505d382bbc);
+    /// @dev Storage slot with the address of the current token deployer. `bytes32(uint256(keccak256('eip1967.proxy.token.deployer')) - 1)`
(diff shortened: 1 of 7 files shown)


src/AxelarGatewayProxy.sol
```diff
@@ -12,9 +12,17 @@ contract AxelarGatewayProxy is EternalStorage {
     /// @dev Storage slot with the address of the current factory. `keccak256('eip1967.proxy.implementation') - 1`.
     bytes32 internal constant KEY_IMPLEMENTATION =
         bytes32(0x360894a13ba1a3210667c828492db98dca3e2076cc3735a920a3ca505d382bbc);
-
-    constructor(address gatewayImplementation, bytes memory params) {
+    /// @dev Storage slot with the address of the current token deployer. `bytes32(uint256(keccak256('eip1967.proxy.token.deployer')) - 1)`
```
sammy1991106, line 15: I don't like the idea of having proxy knowing how implementation is done. Specifically here, we are making the proxy aware that there's such a thing as token deployer which is what the implementation cares about. The proxy should only deal with implementation instead of getting involved into this kind of implementation details.

sammy1991106 (changes requested):

re1ro:
Closing due the contract size bloat

PR #87, opened 2022-04-14
Gas receiver

Foivos:
* [x] Adding the gas receiver files. These enable users to pay gas for contract calls on the source chain, using any token.
* [x] Test coverage for the gas receiver 
* [x] Adding full GMP example + test flow


re1ro (approved):

src/gas-receiver/AxelarGasReceiverProxy.sol
```diff
@@ -0,0 +1,58 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity 0.8.9;
+
+import { AxelarGasReceiver } from './AxelarGasReceiver.sol';
+
+contract AxelarGasReceiverProxy {
```
sammy1991106, line 7: With this proxy pattern, we cannot possibly upgrade the state layout in the implementation at all. Please consider one of those patterns from https://eip2535diamonds.substack.com/p/understanding-delegatecall-and-how?s=r.

src/gas-receiver/AxelarGasReceiver.sol
```diff
@@ -0,0 +1,109 @@
...
+            destinationChain,
+            destinationAddress,
+            payload,
+            symbol,
+            amountThrough,
+            msg.value
+        );
+    }
+
+    function retrieveFees(address payable receiver, address[] memory tokens) external onlyOwner {
```
sammy1991106, line 84:
```suggestion
    function collectFees(address payable receiver, address[] memory tokens) external onlyOwner {
```

src/gas-receiver/AxelarGasReceiver.sol
```diff
@@ -0,0 +1,109 @@
...
+import { IAxelarGasReceiver } from '../interfaces/IAxelarGasReceiver.sol';
+
+// This should be owned by the microservice that is paying for gas.
+contract AxelarGasReceiver is IAxelarGasReceiver, Ownable {
+    bytes32 internal constant _IMPLEMENTATION_SLOT = 0x360894a13ba1a3210667c828492db98dca3e2076cc3735a920a3ca505d382bbc;
+    // solhint-disable-next-line no-empty-blocks
+    constructor() Ownable() {}
+
+    function setup(bytes calldata data) public {
+        (owner) = abi.decode(data, (address));
```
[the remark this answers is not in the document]
Foivos: The constructor is there to satisfy solidity. The proxy does not have the `owner` variable but it does get it once setup is run. It is basically stored somewhere in the memory of the proxy that is correctly accessed by `owner`.

src/gas-receiver/AxelarGasReceiverProxy.sol
Foivos: Will fix this, you are right.

src/gas-receiver/AxelarGasReceiver.sol
Foivos: Sure. Will fix this also.

sammy1991106 (approved):
LGTM but some unit tests would still be great.

src/gas-receiver/AxelarGasReceiver.sol
```diff
@@ -0,0 +1,142 @@
...
+            assembly {
+                sstore(_OWNER_SLOT, newOwner)
+            }
+        }
+    }
+
+    // This is called on the source chain before calling the gateway to execute a remote contract.
+    function payGasForContractCall(
+        address sender,
+        string memory destinationChain,
```
deluca-mike, line 41: In order to not reintroduce issues any c4 brought up, can we set all of these memory variables to calldata?

src/gas-receiver/AxelarGasReceiver.sol
```diff
@@ -0,0 +1,142 @@
...
+        string memory destinationAddress,
+        bytes calldata payload,
+        address gasToken,
+        uint256 gasAmount
+    ) external {
+        IERC20(gasToken).transferFrom(msg.sender, address(this), gasAmount);
```
deluca-mike, line 47: If the gas token is being passed in, and therefore can be anything, we need to use safe transfer functions since not all tokens revert on failure, since some return true or false.

src/gas-receiver/AxelarGasReceiver.sol
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    // This is called on the source chain before calling the gateway to execute a remote contract.
+    function payNativeGasForContractCall(
+        address sender,
+        string memory destinationChain,
+        string memory destinationAddress,
+        bytes calldata payload
+    ) external payable {
+        if (msg.value == 0) revert NothingReceived();
```
deluca-mike, line 89: New lines after if statements, whether they are single line or blocks

src/gas-receiver/AxelarGasReceiver.sol
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    // This is called on the source chain before calling the gateway to execute a remote contract.
+    function payGasForContractCallWithToken(
+        address sender,
+        string memory destinationChain,
+        string memory destinationAddress,
+        bytes calldata payload,
+        string memory symbol,
+        uint256 amountThrough,
```
deluca-mike, line 65: This term is confusingly named.

src/gas-receiver/AxelarGasReceiver.sol
```diff
@@ -0,0 +1,142 @@
...
+            destinationAddress,
+            keccak256(payload),
+            symbol,
+            amountThrough,
+            msg.value
+        );
+    }
+
+    function collectFees(address payable receiver, address[] memory tokens) external onlyOwner {
+        for (uint256 i = 0; i < tokens.length; i++) {
```
deluca-mike, line 115: No need to initialize with zero.

src/gas-receiver/AxelarGasReceiver.sol
```diff
@@ -0,0 +1,142 @@
...
+            if (tokens[i] == address(0)) {
```
deluca-mike, line 116: Should cache the value of tokens[I]

src/gas-receiver/AxelarGasReceiver.sol
```diff
@@ -0,0 +1,142 @@
...
+        _;
+    }
+
+    function owner() public view returns (address owner_) {
+        assembly {
+            owner_ := sload(_OWNER_SLOT)
+        }
+    }
+
+    function setup(bytes calldata data) external override {
```
deluca-mike, line 28: There needs to be a way to prevent calling this function on the implementation itself.

src/gmp-test/ChainASwapCaller.sol
```diff
@@ -0,0 +1,54 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity 0.8.9;
+
+import { IAxelarGateway } from '../interfaces/IAxelarGateway.sol';
+import { IAxelarGasReceiver } from '../interfaces/IAxelarGasReceiver.sol';
+import { IERC20 } from '../interfaces/IERC20.sol';
+
+contract ChainASwapCaller {
```
deluca-mike, line 9: If chain a refers to the source chain and chain b refers to the destination chain, then instead they should be named as such

src/gmp-test/ChainBTokenSwapper.sol
```diff
@@ -4,31 +4,31 @@ pragma solidity 0.8.9;
 
 import { IERC20 } from '../interfaces/IERC20.sol';
 
-contract TokenSwapper {
+contract ChainBTokenSwapper {
```
deluca-mike, line 7: Why can't there be one token swapper on each chain that handles the functionality needed on the source side and destination side?

src/gmp-test/ChainBTokenSwapper.sol
```diff
@@ -4,31 +4,31 @@ pragma solidity 0.8.9;
...
-    address tokenB;
+    address tokenX;
+    address tokenY;
 
-    constructor(address tokenA_, address tokenB_) {
-        tokenA = tokenA_;
-        tokenB = tokenB_;
+    constructor(address tokenX_, address tokenY_) {
+        tokenX = tokenX_;
+        tokenY = tokenY_;
```
deluca-mike, line 15: If x and y correspond to input and output respectively, then they should be named as such.

src/interfaces/IAxelarGasReceiver.sol
```diff
@@ -0,0 +1,102 @@
...
+
+    event GasPaidForContractCall(
+        address sourceAddress,
+        string destinationChain,
+        string destinationAddress,
+        bytes32 payloadHash,
+        address gasToken,
+        uint256 gasAmount
+    );
+    event GasPaidForContractCallWithToken(
```
deluca-mike, line 22: I know it's a nitpick, however there are inconsistencies between new lines between blocks of code. In my opinion, every block of code should be separated by new lines, whether they are multi line event declarations, if blocks, event emissions, etc etc.

deluca-mike (changes requested):

src/gmp-test/ChainBTokenSwapper.sol
re1ro: Token swapper here is the analogy of preexisting Uniswap. Renamed it to `DestinationChainTokenSwapper`

src/gmp-test/ChainBTokenSwapper.sol
re1ro: Renaming back to A and B :)

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@axelar-network/axelar-cgp-solidity",
-  "version": "3.1.1",
+  "version": "3.2.0",
```
sammy1991106, line 3: The version bump should be done separately. This may not be the very last PR for the release.

sammy1991106 (approved):

package.json
re1ro: Will remove that for now

sammy1991106 (approved):

PR #90, opened 2022-04-25
feat: axl token deletion for the sake of re-deployment

sammy1991106:


re1ro (approved):

src/AxelarGatewayMultisig.sol
```diff
@@ -481,6 +481,11 @@ contract AxelarGatewayMultisig is IAxelarGatewayMultisig, AxelarGateway {
 
         emit OwnershipTransferred(new address[](uint256(0)), uint256(0), ownerAddresses, newOwnerThreshold);
         emit OperatorshipTransferred(new address[](uint256(0)), uint256(0), operatorAddresses, newOperatorThreshold);
+
+        // We would need to delete AXL token for the very release and therefore and re-deploy it
+        // with the latest token code that has the potential to support token refund in the future.
+        // This is a one-time thing and should be gotten rid of right after release.
+        _setTokenAddress('AXL', address(0));
```
milapsheth, line 488: Should we add `migrate` function that's run explicitly in the `upgrade` method for future use? This way it doesn't interfere with initial `setup`.

milapsheth (approved):

PR #93, opened 2022-05-02
feat: flatten contracts script

### LINKED ISSUE #85: Generate flattened contracts in a github action
To verify the contracts on the explorers, it would be convenient to generate the flattened contract representations as part of the release build.

To flatten the contracts, you can do:

`npm i @poanet/solidity-flattener`

For every contract that needs to be verified (Proxy, Multisig, Burnable...ERC20, DepositHandler), do

```bash
./node_modules/.bin/poa-solidity-flattener src/BurnableMintableCappedERC20.sol
npm run prettier --write 'out/*.sol'
mv out flattened
```

milapsheth:
Package flattened source code in releases to make contract verification on explorers easier.
fixes #85


package.json
```diff
@@ -36,7 +37,8 @@
     "mocha": "^9.1.3",
     "prettier": "^2.5.1",
     "prettier-plugin-solidity": "^1.0.0-beta.19",
-    "solhint": "^3.3.6"
+    "solhint": "^3.3.6",
+    "@poanet/solidity-flattener": "^3.0.7"
```
sammy1991106, line 41: Why do we need this extra dependency when both hardhat and waffle support flattening? https://coinsbench.com/how-to-flatten-a-solidity-file-using-hardhat-924871ab3139 and https://ethereum-waffle.readthedocs.io/en/latest/getting-started.html

milapsheth: Ah, didn't show up first in my search. I have switched to using waffle. Could switch to hardhat once that's ready.

waffle.json
```diff
@@ -22,5 +22,6 @@
         }
       }
     }
-  }
+  },
+  "flattenOutputDirectory": "./flattened"
```
sammy1991106, line 26: Why not just 
```suggestion
  "flattenOutputDirectory": "./build/flattened"
```

sammy1991106 (approved):

waffle.json
milapsheth: True, updated

PR #97, opened 2022-05-05
Added the ConstAddressDeployer.

Foivos:


re1ro (changes requested):
I think we need some tests for this one. Also how are we passing the constructor params to the deployed contract? I think we need that for our proxies

Foivos:
> Also how are we passing the constructor params to the deployed contract?

Here is some js that uses ethers to deploy this contract and use it to deploy a new contract with some args. 

```js
const deployer = await deployContract(ownerWallet, ConstAddressDeployer);
const salt = keccak256(defaultAbiCoder.encode(['string'], [key]));
const factory = new ContractFactory(
  contractJSON.abi,
  contractJSON.bytecode
);
const bytecode = (await factory.getDeployTransaction(...args)).data;
await (await deployer.connect(wallet).deploy(bytecode, salt)).wait();
const address = await deployer.predict(bytecode, salt);
const contract = new Contract(address, contractJSON.abi, wallet);

```

There is however the following issue: If a contract needs to know params that vary between chains (like the gateway address on each chain) then those cannot be a part of the constructor, because that would cause a different bytecode and a different address as a result. For this reason contract wishing to have the same address. need to specify an empty constructor and an `init` function which will only be able to be called once to setup those variables after deployment.

src/ConstAddressDeployer.sol
```diff
@@ -0,0 +1,23 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity 0.8.9;
+
+contract ConstAddressDeployer {
+    event Deployed(bytes32 bytecodeHash, bytes32 salt, address deployedAddress);
+    function deploy(bytes memory bytecode, bytes32 salt) external returns (address deployedAddress) {
+        // solhint-disable-next-line no-inline-assembly
+        assembly {
+            deployedAddress := create2(0, add(bytecode, 32), mload(bytecode), salt)
```
sammy1991106, line 10: Just to confirm, `add(bytecode, 32)` to skip its length?

src/ConstAddressDeployer.sol
```diff
@@ -0,0 +1,23 @@
...
+        }
+        emit Deployed(keccak256(bytecode), salt, deployedAddress);
+    }
+
+    function predict(bytes memory bytecode, bytes32 salt) external view returns (address deployedAddress) {
```
sammy1991106, line 15:
```suggestion
    function deployedAddress(bytes calldata bytecode, bytes32 salt) external view returns (address deployedAddress) {
```

src/ConstAddressDeployer.sol
```diff
@@ -0,0 +1,23 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity 0.8.9;
+
+contract ConstAddressDeployer {
+    event Deployed(bytes32 bytecodeHash, bytes32 salt, address deployedAddress);
+    function deploy(bytes memory bytecode, bytes32 salt) external returns (address deployedAddress) {
```
sammy1991106, line 7: Can bytecode reside in calldata here or it won't work with mload below?

sammy1991106, line 7:
```suggestion
    event Deployed(bytes32 bytecodeHash, bytes32 salt, address deployedAddress);
    
    function deploy(bytes memory bytecode, bytes32 salt) external returns (address deployedAddress) {
```

sammy1991106: Add an empty line between events and functions.

test/ConstAddressDeployer.js
```diff
@@ -0,0 +1,135 @@
...
+      key, 
+      [name, symbol, decimals, capacity],
+    );
+    expect(await contract.address).to.equal(address);
+    expect(await contract.name()).to.equal(name);
+    expect(await contract.symbol()).to.equal(symbol);
+    expect(await contract.decimals()).to.equal(decimals);
+    expect(await contract.cap()).to.equal(capacity);
+  });
+  it('should deploy to the predicted address even with a different nonce', async () => {
```
sammy1991106, line 80: Add an empty line between tests.

test/ConstAddressDeployer.js
```diff
@@ -0,0 +1,135 @@
...
+        to: userWallet.address,
+        value: 0,
+    });
+    expect(await contract.address).to.equal(address);
+    expect(await contract.name()).to.equal(name);
+    expect(await contract.symbol()).to.equal(symbol);
+    expect(await contract.decimals()).to.equal(decimals);
+    expect(await contract.cap()).to.equal(capacity);
+  });
+  it('should deploy the same contract twice to different addresses with different salts', async () => {
```
sammy1991106, line 106: Same here.

test/ConstAddressDeployer.js
```diff
@@ -0,0 +1,135 @@
...
+}
+
+const deployContractConstant = async (deployer, wallet, contract, key, args = []) => {
+  const salt = getSaltFromKey(key);
+  const factory = new ContractFactory(
+      contract.abi,
+      contract.bytecode,
+  );
+  const bytecode = (await factory.getDeployTransaction(...args)).data;
+  await (await deployer.connect(wallet).deploy(bytecode, salt)).wait();
```
sammy1991106, line 30: It's really hard to read when multiple `await`s exist in one statement.

sammy1991106 (changes requested):

src/ConstAddressDeployer.sol
Foivos: yes, the next argument (mload(bytecode)) is the length.

src/ConstAddressDeployer.sol
Foivos: I think it needs to be memory.

re1ro (approved):

src/test/BurnableMintableCappedERC20Init.sol
```diff
@@ -0,0 +1,74 @@
...
+
+pragma solidity 0.8.9;
+
+import { IAxelarGateway } from '../interfaces/IAxelarGateway.sol';
+import { IBurnableMintableCappedERC20 } from '../interfaces/IBurnableMintableCappedERC20.sol';
+
+import { MintableCappedERC20 } from '../MintableCappedERC20.sol';
+import { DepositHandler } from '../DepositHandler.sol';
+
+contract BurnableMintableCappedERC20Init is IBurnableMintableCappedERC20, MintableCappedERC20 {
```
re1ro, line 11: Would it be possible just to extend `BurnableMintableCappedERC20` and define only constructor and init?

Foivos: Done

Foivos: done.

PR #102, opened 2022-05-24
Adding an option for apps to 'forecall' callContractWithToken to hasten resolution of calls.

Foivos:
Add a different version of the executable that can have executeWithToken front-run by the application (and have the tokens provided to it) to hasten execution.

### DIFF
--- a/contracts/test/gmp/DestinationChainSwapForecallable.sol
+++ b/contracts/test/gmp/DestinationChainSwapForecallable.sol
@@ -9,7 +9,7 @@ import { DestinationChainTokenSwapper } from './DestinationChainTokenSwapper.sol
 contract DestinationChainSwapForecallable is IAxelarForecallable {
     DestinationChainTokenSwapper public swapper;
 
-    constructor(address gatewayAddress, address swapperAddress) IAxelarExecutableForecallable(gatewayAddress) {
+    constructor(address gatewayAddress, address swapperAddress) IAxelarForecallable(gatewayAddress) {
         swapper = DestinationChainTokenSwapper(swapperAddress);
     }
 


sammy1991106:
I need more information about what is this PR doing and what are we trying to achieve before I can review it.

test/Util.js
```diff
@@ -0,0 +1,31 @@
+'use strict';
+
+const chai = require('chai');
+const { deployContract, MockProvider, solidity } = require('ethereum-waffle');
+chai.use(solidity);
+const { expect } = chai;
+
+const UtilTest = require('../build/UtilTest.json');
+
+const { it } = require('mocha');
```
sammy1991106, line 10:
```suggestion
```

src/interfaces/IAxelarExecutableForetellable.sol
```diff
@@ -0,0 +1,101 @@
...
+import { IAxelarGateway } from './IAxelarGateway.sol';
+import { IERC20 } from './IERC20.sol';
+
+abstract contract IAxelarExecutableForetellable {
+    error NotApprovedByGateway();
+    error AlreadyFortold();
+    error NotAuthorizedToForetell();
+
+    IAxelarGateway public gateway;
+    mapping(string => mapping(string => mapping(bytes => mapping(string => mapping(uint256 => address)))))
```
sammy1991106, line 14:
```suggestion
    mapping(bytes32 => address) public foretellers;
```
and key becomes keccak256(sourceChain, sourceAddress, payload, tokenSymbol, amount). Fewer hashing happening and therefore cheaper gas cost, no?

src/gmp-test/DestinationChainSwapExecutableForetellable.sol
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    function _executeWithToken(
+        string memory sourceChain,
+        string memory,
+        bytes calldata payload,
+        string memory tokenSymbolA,
+        uint256 amount
+    ) internal override {
+        (string memory tokenSymbolB, string memory recipient) = abi.decode(payload, (string, string));
```
sammy1991106, line 23:
```suggestion
        (bytes32 nonce, string memory tokenSymbolB, string memory recipient) = abi.decode(payload, (string, string));
```
Without a nonce, only one of the multiple swapping attempts with the exact same parameters can be foretold until executed.

src/interfaces/IAxelarExecutableForetellable.sol
```diff
@@ -0,0 +1,101 @@
...
+        string calldata sourceAddress,
+        bytes calldata payload
+    ) external {
+        bytes32 payloadHash = keccak256(payload);
+        if (!gateway.validateContractCall(commandId, sourceChain, sourceAddress, payloadHash))
+            revert NotApprovedByGateway();
+        _execute(sourceChain, sourceAddress, payload);
+    }
+
+    function foretell(
```
sammy1991106, line 33: I don't see why foretell can only work with executeWithToken but not execute?

sammy1991106 (dismissed):

re1ro:
Please rebase your branch onto the latest main 💯

src/interfaces/IAxelarExecutableForetellable.sol
```diff
@@ -0,0 +1,101 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity 0.8.9;
+
+import { IAxelarGateway } from './IAxelarGateway.sol';
+import { IERC20 } from './IERC20.sol';
+
+abstract contract IAxelarExecutableForetellable {
```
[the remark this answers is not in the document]
re1ro: @sammy1991106 we already used that pattern for the `IAxelarExecutable`. It has some logic in it and also is an abstract contract. Do you think it's worth renaming it? Also should it be moved out of `interfaces` folder?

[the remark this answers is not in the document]
sammy1991106: ok nvm then if it's already done elsewhere.

contracts/interfaces/IAxelarExecutableForetellable.sol
```diff
@@ -0,0 +1,118 @@
...
+    ) external {
+        bytes32 payloadHash = keccak256(payload);
+        if (!gateway.validateContractCall(commandId, sourceChain, sourceAddress, payloadHash)) revert NotApprovedByGateway();
+        address foreteller = foretellers[keccak256(abi.encode(sourceChain, sourceAddress, payload))];
+        if (foreteller != address(0)) {
+            foretellers[keccak256(abi.encode(sourceChain, sourceAddress, payload))] = address(0);
+        } else {
+            _execute(sourceChain, sourceAddress, payload);
+        }
+        _execute(sourceChain, sourceAddress, payload);
```
re1ro, line 46: Is this `_execute` duplicated here?

contracts/interfaces/IAxelarExecutableForetellable.sol
```diff
@@ -0,0 +1,118 @@
...
+    function foretellWithToken(
+        string calldata sourceChain,
+        string calldata sourceAddress,
+        bytes calldata payload,
+        string calldata tokenSymbol,
+        uint256 amount,
+        address foreteller
+    ) external {
+        address token = gateway.tokenAddresses(tokenSymbol);
+        IERC20(token).transferFrom(msg.sender, address(this), amount);
```
re1ro, line 58: We need to use safe transfer here
[Example](https://github.com/axelarnetwork/axelar-cgp-solidity/blob/main/contracts/gas-service/AxelarGasService.sol#L158)

re1ro (changes requested):

re1ro (approved):

PR #104, opened 2022-05-27
Refactoring deposit service and IERC20 interface

re1ro:
* [x] DepositService: renaming nonce to salt
* [x] DepositService: exposing hashing keys through the interface
* [x] DepositService: testing deposit address generation
* [x] IERC20: adding name, symbol and decimals to the interface

### DIFF
--- a/src/ERC20.sol
+++ b/src/ERC20.sol
@@ -35,10 +35,10 @@ contract ERC20 is IERC20 {
 
     uint256 public override totalSupply;
 
-    string public override name;
-    string public override symbol;
+    string public name;
+    string public symbol;
 
-    uint8 public immutable override decimals;
+    uint8 public immutable decimals;
 
     /**
      * @dev Sets the values for {name}, {symbol}, and {decimals}.
--- a/src/ERC20.sol
+++ b/src/ERC20.sol
@@ -35,10 +35,10 @@ contract ERC20 is IERC20 {
 
     uint256 public override totalSupply;
 
-    string public name;
-    string public symbol;
+    string public override name;
+    string public override symbol;
 
-    uint8 public immutable decimals;
+    uint8 public immutable override decimals;
 
     /**
      * @dev Sets the values for {name}, {symbol}, and {decimals}.
--- a/src/deposit-service/AxelarDepositService.sol
+++ b/src/deposit-service/AxelarDepositService.sol
@@ -17,9 +17,9 @@ contract AxelarDepositService is Upgradable, IAxelarDepositService {
     bytes32 internal constant _WRAPPED_TOKEN_SYMBOL_SLOT =
         0x91d2f5305ae2a8f5b319f6c3a690eff002c3e572220774ba5f7e957f079e55df;
 
-    bytes32 public constant override PREFIX_DEPOSIT_SEND_TOKEN = keccak256('deposit-send-token');
-    bytes32 public constant override PREFIX_DEPOSIT_SEND_NATIVE = keccak256('deposit-send-native');
-    bytes32 public constant override PREFIX_DEPOSIT_WITHDRAW_NATIVE = keccak256('deposit-withdraw-native');
+    bytes32 internal constant PREFIX_DEPOSIT_SEND_TOKEN = keccak256('deposit-send-token');
+    bytes32 internal constant PREFIX_DEPOSIT_SEND_NATIVE = keccak256('deposit-send-native');
+    bytes32 internal constant PREFIX_DEPOSIT_WITHDRAW_NATIVE = keccak256('deposit-withdraw-native');
 
     function depositAddressForSendToken(
         bytes32 salt,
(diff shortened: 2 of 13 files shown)


src/ERC20.sol
```diff
@@ -35,10 +35,10 @@ contract ERC20 is IERC20 {
 
     uint256 public override totalSupply;
 
-    string public name;
-    string public symbol;
+    string public override name;
```
milapsheth, line 38: Now that we have split the token deployment from the gateway, I think it's worth having proper methods for these for clarity. Would the increased bytecode size have a significant gas impact?

Otherwise, can you add to the doc above that allowance etc. implement the IERC20 interface because of how variables are referred to via public methods?

src/interfaces/IERC20.sol
```diff
@@ -36,6 +36,21 @@ interface IERC20 {
...
+
+    /**
+     * @dev Returns the symbol of the token.
+     */
+    function symbol() external view returns (string memory);
+
+    /**
+     * @dev Returns the amount of decimal points for displaying a token amount.
+     */
+    function decimals() external view returns (uint8);
```
milapsheth, line 52: So, it seems that `IERC20` doesn't require these methods. But, there is `IERC20Metadata` that adds these extensions. Let's [mimic](https://docs.openzeppelin.com/contracts/4.x/api/token/erc20#IERC20Metadata) having two separate interfaces that our ERC20 contracts implement? This way we don't assume that these [optional methods](https://eips.ethereum.org/EIPS/eip-20#methods) are required for external tokens.

test/AxelarDepositService.js
```diff
@@ -206,16 +251,33 @@ describe('AxelarDepositService', () => {
...
+
+      expect(depositAddress).to.be.equal(expectedDepositAddress);
 
       await token.connect(ownerWallet).transfer(depositAddress, amount);
 
       await expect(
-        await depositService.withdrawNative(nonce, recipient),
+        await depositService.withdrawNative(salt, recipient),
       ).to.changeEtherBalance(userWallet, amount);
     });
```
milapsheth, line 282: Maybe add some tests for name/symbol etc, if we don't have those?

src/ERC20.sol
sammy1991106: "Otherwise, can you add to the doc above that allowance etc. implement the IERC20 interface because of how variables are referred to via public methods?" I don't think that's necessary since it's a common knowledge of solidity.

src/ERC20.sol
```diff
@@ -35,10 +35,10 @@ contract ERC20 is IERC20 {
...
+    string public override symbol;
 
-    uint8 public immutable decimals;
+    uint8 public immutable override decimals;
```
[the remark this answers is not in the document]
sammy1991106: When you verify, you'd have to pass in the constructor param so the `immutable` here doesn't make anything harder.

src/deposit-service/AxelarDepositService.sol
```diff
@@ -17,32 +17,32 @@ contract AxelarDepositService is Upgradable, IAxelarDepositService {
     bytes32 internal constant _WRAPPED_TOKEN_SYMBOL_SLOT =
         0x91d2f5305ae2a8f5b319f6c3a690eff002c3e572220774ba5f7e957f079e55df;
 
-    bytes32 internal constant PREFIX_DEPOSIT_SEND_TOKEN = keccak256('deposit-send-token');
-    bytes32 internal constant PREFIX_DEPOSIT_SEND_NATIVE = keccak256('deposit-send-native');
-    bytes32 internal constant PREFIX_DEPOSIT_WITHDRAW_NATIVE = keccak256('deposit-withdraw-native');
+    bytes32 public constant override PREFIX_DEPOSIT_SEND_TOKEN = keccak256('deposit-send-token');
```
sammy1991106, line 20: Why do we want to expose these?

src/interfaces/IERC20.sol
sammy1991106: To be fair, these are all optional in the ERC20 standard so none of these should be part of `IERC20`.

src/deposit-service/AxelarDepositService.sol
re1ro: For transparency, so the deposit address can be calculated offline. Do you think it's not worth it and those constants could be found in the source code?

src/interfaces/IERC20.sol
re1ro: Got it. I will remove the extras then

test/AxelarDepositService.js
re1ro: We are removing `IERC20Metadata` from our IERC20 interface. No need to test those

sammy1991106 (approved):

src/deposit-service/AxelarDepositService.sol
re1ro: Removing it as we agreed

PR #105, opened 2022-06-01
feat/freeze removal

sammy1991106:


src/AxelarGateway.sol
```diff
@@ -43,12 +42,9 @@ abstract contract AxelarGateway is IAxelarGateway, AdminMultisigBase {
         bytes32(0x360894a13ba1a3210667c828492db98dca3e2076cc3735a920a3ca505d382bbc);
 
     // AUDIT: slot names should be prefixed with some standard string
-    bytes32 internal constant KEY_ALL_TOKENS_FROZEN = keccak256('all-tokens-frozen');
-
     bytes32 internal constant PREFIX_COMMAND_EXECUTED = keccak256('command-executed');
     bytes32 internal constant PREFIX_TOKEN_ADDRESS = keccak256('token-address');
     bytes32 internal constant PREFIX_TOKEN_TYPE = keccak256('token-type');
-    bytes32 internal constant PREFIX_TOKEN_FROZEN = keccak256('token-frozen');
```
milapsheth, line 51: Can we keep the prefixes either commented out or some other way so we don't have prefix collisions in the future?

re1ro (approved):

PR #108, opened 2022-06-03
chore(README): updating dependencies and build instructions

re1ro:
Updating README to reflect latest hardhat changes

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,6 +43,6 @@
     "solidity-coverage": "^0.7.17"
   },
   "engines": {
-    "node": ">=14"
+    "node": ">=16.15.1 <17"
   }
 }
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -43,6 +43,6 @@
     "solidity-coverage": "^0.7.17"
   },
   "engines": {
-    "node": ">=14"
+    "node": ">=16.15.1 <17"
```
milapsheth, line 46: Using the latest v16 release seems too restrictive since most people don't update for every patch?
```suggestion
    "node": ">=16 <17"
```

milapsheth:
Maybe bump up the dev dependencies too?

package.json
sammy1991106: Agree.

milapsheth (approved):

Foivos (approved):

PR #114, opened 2022-06-17
Allow for forcall fee to be part of the payload.

Foivos:


sammy1991106 (approved):

re1ro:
@Foivos does it make sense to write some tests for it? So they can serve as an example for using forecall with fee

Foivos:
> @Foivos does it make sense to write some tests for it? So they can serve as an example for using forecall with fee

I am making a full fledged example in the examples repo rn.

PR #119, opened 2022-06-22
reduce to ^0.8.0

Foivos:


re1ro (changes requested):
Could you change this to `^0.8.9;` to be a bit more restrictive but support latest versions?

re1ro (approved):

PR #127, opened 2022-07-18
fix: check non-zero amount for transfer

milapsheth:


contracts/gas-service/AxelarGasService.sol
```diff
@@ -149,6 +149,8 @@ contract AxelarGasService is Upgradable, IAxelarGasService {
         address receiver,
         uint256 amount
     ) internal {
+        if (amount == 0) revert NothingReceived();
```
milapsheth, line 152: Revert on `0` amount similar to the payNativeGas method.

re1ro (approved):

PR #136, opened 2022-08-05
add doc for function allTokensFrozen

sammy1991106:


contracts/AxelarGateway.sol
```diff
@@ -157,6 +157,10 @@ contract AxelarGateway is IAxelarGateway, AdminMultisigBase {
         return getUint(_getTokenDailyMintAmountKey(symbol, block.timestamp / 1 days));
     }
 
+    /*
+     * @dev the function is kept around to keep things work for tokens that were
```
milapsheth, line 161:
```suggestion
     * @dev This function is kept around to keep things working for internal tokens that were
```
Also, add the doc for `tokenFrozen`?

milapsheth (approved):

re1ro (approved):

PR #137, opened 2022-08-08
fix(Gateway): addressing the CodeArena audit

re1ro:


contracts/AxelarGateway.sol
```diff
@@ -454,22 +460,10 @@ contract AxelarGateway is IAxelarGateway, AdminMultisigBase {
...
-        )
-    {
-        (chainId, , commandIds, commands, params) = abi.decode(executeData, (uint256, uint256, bytes32[], string[], bytes[]));
-    }
-
     function _callERC20Token(address tokenAddress, bytes memory callData) internal returns (bool) {
+        // solhint-disable-next-line avoid-low-level-calls
         (bool success, bytes memory returnData) = tokenAddress.call(callData);
-        return success && (returnData.length == uint256(0) || abi.decode(returnData, (bool)));
+        return success && tokenAddress.code.length != 0 && (returnData.length == 0 || abi.decode(returnData, (bool)));
```
milapsheth, line 466: I suppose this check is useful for external tokens that might selfdestruct for some reason?

contracts/AxelarGateway.sol
```diff
@@ -551,7 +545,8 @@ contract AxelarGateway is IAxelarGateway, AdminMultisigBase {
     }
 
     function _getTokenDailyMintAmountKey(string memory symbol, uint256 day) internal pure returns (bytes32) {
-        return keccak256(abi.encodePacked(PREFIX_TOKEN_DAILY_MINT_AMOUNT, symbol, day));
+        // abi.encode to securely hash dynamic-length symbol data followed by day
+        return keccak256(abi.encode(PREFIX_TOKEN_DAILY_MINT_AMOUNT, symbol, day));
```
milapsheth, line 549: Nice, alternatively, we encode the fixed length day first followed by symbol with `encodePacked` instead

contracts/ERC20.sol
```diff
@@ -169,8 +169,8 @@ contract ERC20 is IERC20 {
 
         _beforeTokenTransfer(sender, recipient, amount);
 
-        balanceOf[sender] -= amount;
-        balanceOf[recipient] += amount;
+        balanceOf[sender] = balanceOf[sender] - amount;
```
milapsheth, line 172: Why is this cheaper?

contracts/ERC20.sol
```diff
@@ -209,8 +209,8 @@ contract ERC20 is IERC20 {
 
         _beforeTokenTransfer(account, address(0), amount);
 
-        balanceOf[account] -= amount;
-        totalSupply -= amount;
+        balanceOf[account] = balanceOf[account] - amount;
+        totalSupply = totalSupply + amount;
```
milapsheth, line 213:
```suggestion
        totalSupply = totalSupply - amount;
```
Should add unit tests to catch this.

contracts/ERC20.sol
```diff
@@ -209,8 +209,8 @@ contract ERC20 is IERC20 {
 
         _beforeTokenTransfer(account, address(0), amount);
```
milapsheth, line 210: Let's remove this method now? No need to have newly deployed tokens waste gas with this

contracts/auth/AxelarAuthWeighted.sol
```diff
@@ -37,7 +43,8 @@ contract AxelarAuthWeighted is Ownable, IAxelarAuthWeighted {
 
         _validateSignatures(messageHash, operators, weights, threshold, signatures);
 
-        currentOperators = operatorsEpoch == epoch;
+        // Returns true for the current operators
```
milapsheth, line 46: Add natspec doc for this instead

contracts/auth/AxelarAuthWeighted.sol
```diff
@@ -65,15 +72,18 @@ contract AxelarAuthWeighted is Ownable, IAxelarAuthWeighted {
...
+            unchecked {
+                ++i;
+            }
         }
         if (newThreshold == 0 || totalWeight < newThreshold) revert InvalidThreshold();
 
         bytes32 newOperatorsHash = keccak256(params);
 
-        if (epochForHash[newOperatorsHash] > 0) revert SameOperators();
+        if (epochForHash[newOperatorsHash] != 0) revert SameOperators();
```
milapsheth, line 86: `SameOperators` -> `DuplicateOperatorCommitment` maybe?

milapsheth (changes requested):

contracts/AxelarGateway.sol
```diff
@@ -154,14 +156,12 @@ contract AxelarGateway is IAxelarGateway, AdminMultisigBase {
...
         return getUint(_getTokenDailyMintAmountKey(symbol, block.timestamp / 1 days));
     }
 
-    /*
-     * @dev This function is kept around to keep things working for internal
-     * tokens that were deployed before the token freeze functionality was
-     * removed
-     */
+    // @dev This function is kept around to keep things working for internal
+    // tokens that were deployed before the token freeze functionality was removed
```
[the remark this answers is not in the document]
sammy1991106: Both `///` and `/*` follow the instruction of NatSpec format https://docs.soliditylang.org/en/v0.8.15/natspec-format.html#natspec-format, but definitely not `//`.

contracts/AdminMultisigBase.sol
```diff
@@ -48,8 +48,11 @@ contract AdminMultisigBase is EternalStorage {
 
         uint256 adminCount = _getAdminCount(adminEpoch);
 
-        for (uint256 i; i < adminCount; ++i) {
+        for (uint256 i; i != adminCount; ) {
             _setHasVoted(adminEpoch, topic, _getAdmin(adminEpoch, i), false);
+            unchecked {
```
sammy1991106, line 53: I don't see any good in this style, and it hurts safety.

contracts/AxelarGateway.sol
```diff
@@ -42,8 +42,10 @@ contract AxelarGateway is IAxelarGateway, AdminMultisigBase {
     bytes32 internal constant SELECTOR_APPROVE_CONTRACT_CALL_WITH_MINT = keccak256('approveContractCallWithMint');
     bytes32 internal constant SELECTOR_TRANSFER_OPERATORSHIP = keccak256('transferOperatorship');
 
-    address internal immutable AUTH_MODULE;
-    address internal immutable TOKEN_DEPLOYER_IMPLEMENTATION;
+    // solhint-disable-next-line var-name-mixedcase
+    address public immutable AUTH_MODULE;
+    // solhint-disable-next-line var-name-mixedcase
+    address public immutable TOKEN_DEPLOYER_IMPLEMENTATION;
```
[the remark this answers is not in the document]
sammy1991106: Agree with @milapsheth here. PRs as big as this is too hard to review.

contracts/AxelarGateway.sol
```diff
@@ -202,40 +199,51 @@ contract AxelarGateway is IAxelarGateway, AdminMultisigBase {
...
+            }
         }
     }
 
     function upgrade(
         address newImplementation,
         bytes32 newImplementationCodeHash,
         bytes calldata setupParams
-    ) external override onlyAdmin {
+    ) external payable override onlyAdmin {
```
sammy1991106, line 238: Why payable

contracts/AxelarGateway.sol
```diff
@@ -282,24 +291,15 @@ contract AxelarGateway is IAxelarGateway, AdminMultisigBase {
...
-            uint256 chainId_,
-            bytes32[] memory commandIds_,
-            string[] memory commands_,
-            bytes[] memory params_
-        ) {
-            (chainId, commandIds, commands, params) = (chainId_, commandIds_, commands_, params_);
-        } catch {
-            (chainId, commandIds, commands, params) = abi.decode(data, (uint256, bytes32[], string[], bytes[]));
-        }
+        (chainId, commandIds, commands, params) = abi.decode(data, (uint256, bytes32[], string[], bytes[]));
```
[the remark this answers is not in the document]
sammy1991106: I am not sure this is a good idea. What's our procedure to make sure all legacy batches are submitted before upgrade?

contracts/AxelarGateway.sol
```diff
@@ -327,18 +327,23 @@ contract AxelarGateway is IAxelarGateway, AdminMultisigBase {
...
+            } else _setCommandExecuted(commandId, false);
         }
     }
 
     /******************\
     |* Self Functions *|
     \******************/
 
-    function deployToken(bytes calldata params, bytes32) external onlySelf {
+    function deployToken(bytes calldata params, bytes32) external payable onlySelf {
```
sammy1991106, line 346: Why payable

contracts/AxelarGateway.sol
sammy1991106: Why even bother calling if `tokenAddress.code.length` is zero

contracts/gas-service/AxelarGasService.sol
```diff
@@ -4,7 +4,7 @@ pragma solidity 0.8.9;
```
sammy1991106, line 4: Please don't change gateway and then gas service in the same PR in the future. PRs need to be focused in terms of the context.

contracts/gas-service/AxelarGasService.sol
```diff
@@ -134,14 +138,20 @@ contract AxelarGasService is Upgradable, IAxelarGasService {
...
         uint256 amount
-    ) external onlyOwner {
+    ) external payable onlyOwner {
         if (receiver == address(0)) revert InvalidAddress();
 
         if (token == address(0)) {
-            receiver.transfer(amount);
+            if (amount <= address(this).balance) {
+                // solhint-disable-next-line avoid-low-level-calls
+                (bool sent, ) = receiver.call{ value: amount }('');
```
sammy1991106, line 150: Why is this better? `.transfer` only forwards 2300 gas as a stipend. Once using `.call`, the security implication is completely different as all gas is being forwarded.

sammy1991106 (changes requested):

contracts/AxelarGateway.sol
Foivos: it saves like 21 gas because some checks are not run when calling this but I don't think that's enough justification for the confusion it adds.

contracts/AxelarGateway.sol
Foivos: Again, saves some gas, again, don't think it matters enough.

contracts/AxelarGateway.sol
Foivos: I agree with Sammy, this check makes more sense before the call.

contracts/AxelarGateway.sol
Foivos: Since there is only one unfixed-length variable I don't think it matters actually: The old encodePacked was always 32b followed by an unknown length, followed by another 32b, Which is unique. If you have more than one multiple length variables however then thery can be manipulated to create the same encoding from different inputs.

contracts/gas-service/AxelarGasService.sol
Foivos: It is possible the `receiver` is a multisig contract that does something fancy on `receive` so 2300 gas is not enough. But you are right that the security implications are a lot more complicated but I can't think of any vulnerabilities (re-entrancy is covered by onlyOwner and memory is not used). If you can come up with something lmk, but even if we can't think of something i'm not sure if it's worh the risk just to support weird multisigs.

sammy1991106: @Foivos  One of those things people do is to check if there's any excessive gas provided, and if so the receiver contract would use as much as possible to get the gas token. In general, the 2300 stipend comes with a security reason that no SSTORE can be performed. I'd prefer not to provide any more gas other than that unless there's a use case already.

contracts/AxelarGateway.sol
sammy1991106: agree. I don't think it justifies at all if all we are doing here is saving 21 gas.

contracts/AxelarGateway.sol
milapsheth: Ah, yes, although `encode` is just safer to reason about, so I'm ok with switching to it

re1ro:
Closing as we are not going to implement all the minor gas optimizations

PR #138, opened 2022-08-08
fix(Gateway): addressing the important CodeArena reports

re1ro:
* [x] Setting `currentOperators` to false when transfer is command successfully executed
* [x] Removed support for legacy batch with signer role. Fixed tests using legacy batches
* [x] Fix the abi encoding of the key for `dailyMintAmount`
* [x] Additional security unit tests for the gateway
* [x] Made auth module/token deployer public and queryable.
* [x] Code length check for `_callERC20Token`
* [x] Suppressed lint warnings with comments

### DIFF
--- a/contracts/AxelarGateway.sol
+++ b/contracts/AxelarGateway.sol
@@ -453,7 +453,10 @@ contract AxelarGateway is IAxelarGateway, AdminMultisigBase {
     function _callERC20Token(address tokenAddress, bytes memory callData) internal returns (bool) {
         // solhint-disable-next-line avoid-low-level-calls
         (bool success, bytes memory returnData) = tokenAddress.call(callData);
-        return success && tokenAddress.code.length != 0 && (returnData.length == 0 || abi.decode(returnData, (bool)));
+        if (tokenAddress.code.length == 0) return false
+
+        (bool success, bytes memory returnData) = tokenAddress.call(callData);
+        return success && (returnData.length == 0 || abi.decode(returnData, (bool)));
     }
 
     function _mintToken(
--- a/contracts/AxelarGateway.sol
+++ b/contracts/AxelarGateway.sol
@@ -453,10 +453,7 @@ contract AxelarGateway is IAxelarGateway, AdminMultisigBase {
     function _callERC20Token(address tokenAddress, bytes memory callData) internal returns (bool) {
         // solhint-disable-next-line avoid-low-level-calls
         (bool success, bytes memory returnData) = tokenAddress.call(callData);
-        if (tokenAddress.code.length == 0) return false
-
-        (bool success, bytes memory returnData) = tokenAddress.call(callData);
-        return success && (returnData.length == 0 || abi.decode(returnData, (bool)));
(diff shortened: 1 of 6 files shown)


contracts/AxelarGateway.sol
```diff
@@ -454,22 +450,10 @@ contract AxelarGateway is IAxelarGateway, AdminMultisigBase {
...
-        )
-    {
-        (chainId, , commandIds, commands, params) = abi.decode(executeData, (uint256, uint256, bytes32[], string[], bytes[]));
-    }
-
     function _callERC20Token(address tokenAddress, bytes memory callData) internal returns (bool) {
+        // solhint-disable-next-line avoid-low-level-calls
         (bool success, bytes memory returnData) = tokenAddress.call(callData);
-        return success && (returnData.length == uint256(0) || abi.decode(returnData, (bool)));
+        return success && tokenAddress.code.length != 0 && (returnData.length == 0 || abi.decode(returnData, (bool)));
```
sammy1991106, line 456:
```suggestion
        if (tokenAddress.code.length == 0) return false

        (bool success, bytes memory returnData) = tokenAddress.call(callData);
        return success && (returnData.length == 0 || abi.decode(returnData, (bool)));
```

contracts/interfaces/IAxelarGateway.sol
```diff
@@ -144,6 +144,12 @@ interface IAxelarGateway {
     |* Getters *|
     \***********/
 
+    // solhint-disable-next-line func-name-mixedcase
+    function AUTH_MODULE() external view returns (address);
```
sammy1991106, line 148: These function names are terrible. I'd rather keep those two vars internal/private and create getters for them ourselves instead of using the auto-generated ones.

sammy1991106 (changes requested):

contracts/interfaces/IAxelarGateway.sol
milapsheth: Yeah, I prefer that too

contracts/AxelarGateway.sol
```diff
@@ -327,10 +317,15 @@ contract AxelarGateway is IAxelarGateway, AdminMultisigBase {
...
+            // solhint-disable-next-line avoid-low-level-calls
             (bool success, ) = address(this).call(abi.encodeWithSelector(commandSelector, params[i], commandId));
 
-            if (success) emit Executed(commandId);
-            else _setCommandExecuted(commandId, false);
+            if (success) {
+                emit Executed(commandId);
+
+                // Not current operators anymore
+                if (commandSelector == AxelarGateway.transferOperatorship.selector) currentOperators = false;
```
milapsheth, line 327: https://github.com/axelarnetwork/axelar-cgp-solidity/pull/137/files#r939821225

contracts/interfaces/IAxelarGateway.sol
re1ro: Fixed

sammy1991106 (approved):

PR #140, opened 2022-08-08
fix(AuthWeighted): improving gas usage

re1ro:
* [x] Improving gas usage by not reading from an array twice and not declaring default values


milapsheth (approved):

contracts/auth/AxelarAuthWeighted.sol
```diff
@@ -115,12 +119,21 @@ contract AxelarAuthWeighted is Ownable, IAxelarAuthWeighted {
...
+            prevAccount = account;
+            account = accounts[i];
+
+            if (prevAccount >= account) {
                 return false;
             }
         }
 
-        return accounts[0] != address(0);
+        return true;
```
sammy1991106, line 137:
```suggestion
        uint256 accountsLength = accounts.length;
        address prevAccount = accounts[0];

        if (prevAccount == address(0)) return false;

        for (uint256 i = 1; i < accountsLength; ++i) {
            address currAccount = accounts[i];

            if (prevAccount >= currAccount) {
                return false;
            }

            prevAccount = currAccount;
        }

        return true;
```

PR #151, opened 2022-09-07
feat(deposit-serivce): deploy script

jack0son:


re1ro:
@jack0son I am closing this PR in favor of #152 and `deploy-upgradable.js` that we already have for this purpose. Please let me know if there is anything you need to be added there

PR #153, opened 2022-09-23
feat(DepositService): refundIssuer for locked assets

re1ro:
* [x] DepositService: adding refund functionality for locked assets


Foivos (approved):

contracts/deposit-service/AxelarDepositService.sol
```diff
@@ -240,6 +246,21 @@ contract AxelarDepositService is Upgradable, DepositServiceBase, IAxelarDepositS
...
+    function refundLockedAsset(
+        address payable receiver,
+        address token,
+        uint256 amount
+    ) external {
+        if (msg.sender != refundIssuer) revert NotRefundIssuer();
+        if (receiver == address(0)) revert InvalidAddress();
+
+        if (token == address(0)) {
+            receiver.transfer(amount);
```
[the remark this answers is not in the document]
re1ro: Addressed 👍

contracts/deposit-service/AxelarDepositService.sol
```diff
@@ -240,6 +246,21 @@ contract AxelarDepositService is Upgradable, DepositServiceBase, IAxelarDepositS
         refundToken = address(0);
     }
 
+    function refundLockedAsset(
+        address payable receiver,
```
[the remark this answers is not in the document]
re1ro: Addressed 👍

contracts/deposit-service/AxelarDepositService.sol
```diff
@@ -240,6 +246,21 @@ contract AxelarDepositService is Upgradable, DepositServiceBase, IAxelarDepositS
...
+        address token,
+        uint256 amount
+    ) external {
+        if (msg.sender != refundIssuer) revert NotRefundIssuer();
+        if (receiver == address(0)) revert InvalidAddress();
```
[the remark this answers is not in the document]
re1ro: Addressed 👍

contracts/deposit-service/DepositServiceBase.sol
```diff
@@ -14,21 +14,23 @@ contract DepositServiceBase is IDepositServiceBase {
...
 
     constructor(address gateway_, string memory wrappedSymbol_) {
         if (gateway_ == address(0)) revert InvalidAddress();
 
         gateway = gateway_;
+        wrappedTokenAddress = IAxelarGateway(gateway_).tokenAddresses(wrappedSymbol_);
 
         // Checking if token symbol exists in the gateway
-        if (IAxelarGateway(gateway_).tokenAddresses(wrappedSymbol_) == address(0)) revert InvalidSymbol();
+        if (bytes(wrappedSymbol_).length > 0 && wrappedTokenAddress == address(0)) revert InvalidSymbol();
```
milapsheth, line 27:
```suggestion
        if (bytes(wrappedSymbol_).length > 0) {
            wrappedTokenAddress = IAxelarGateway(gateway_).tokenAddresses(wrappedSymbol_);
            if (wrappedTokenAddress == address(0)) revert InvalidSymbol();
        }
```

maybe something like this instead?

contracts/deposit-service/DepositServiceBase.sol
```diff
@@ -14,21 +14,23 @@ contract DepositServiceBase is IDepositServiceBase {
...
     // Using immutable storage to keep the constants in the bytecode
     address public immutable gateway;
+    address public immutable wrappedTokenAddress;
     bytes32 internal immutable wrappedSymbolBytes;
 
     constructor(address gateway_, string memory wrappedSymbol_) {
         if (gateway_ == address(0)) revert InvalidAddress();
 
         gateway = gateway_;
+        wrappedTokenAddress = IAxelarGateway(gateway_).tokenAddresses(wrappedSymbol_);
```
milapsheth, line 24:
```suggestion
```
see below

contracts/deposit-service/DepositServiceBase.sol
re1ro: This pattern throws a compiler error. But I will add the comment to make that line more clear
```
TypeError: Cannot write to immutable here: 
Immutable variables cannot be initialized inside an if statement.
```

milapsheth (approved):

contracts/deposit-service/DepositServiceBase.sol
```diff
@@ -14,21 +14,23 @@ contract DepositServiceBase is IDepositServiceBase {
...
-
-        // Checking if token symbol exists in the gateway
-        if (IAxelarGateway(gateway_).tokenAddresses(wrappedSymbol_) == address(0)) revert InvalidSymbol();
-
+        wrappedTokenAddress = IAxelarGateway(gateway_).tokenAddresses(wrappedSymbol_);
         wrappedSymbolBytes = wrappedSymbol_.toBytes32();
+
+        // Wrapped token symbol param is optional
+        // When specified we are checking if token exists in the gateway
+        if (bytes(wrappedSymbol_).length > 0 && wrappedTokenAddress == address(0)) revert InvalidSymbol();
```
sammy1991106, line 29:
```suggestion
        if (wrappedSymbol_.length == 0) return;

        wrappedTokenAddress = IAxelarGateway(gateway_).tokenAddresses(wrappedSymbol_);

        if (wrappedTokenAddress == address(0)) revert InvalidSymbol();

        wrappedSymbolBytes = wrappedSymbol_.toBytes32();
```

contracts/deposit-service/AxelarDepositService.sol
```diff
@@ -240,6 +248,25 @@ contract AxelarDepositService is Upgradable, DepositServiceBase, IAxelarDepositS
         refundToken = address(0);
     }
 
+    function refundLockedAsset(
```
sammy1991106, line 251: All good but it feels a little scary that the refund issuer has so much power that it can send however amount to wherever it wants to.

sammy1991106 (changes requested):

contracts/deposit-service/AxelarDepositService.sol
milapsheth: since it's unlikely that we would need to refund from the deposit service often, maybe we should only let the owner collect the funds? @re1ro

milapsheth: ok, yeah, initially a bunch of apps won't set the refund address (since it'll take some time to work it in the UI), so I can see it being useful to have it be a hot wallet instead. and we can swap that out later to a cold wallet if needed.

Foivos (approved):

contracts/deposit-service/DepositServiceBase.sol
re1ro: @sammy1991106 in this PR we are allowing the empty `wrappedSymbol_ ` for the chains where we don't support their wrapped token (Aurora etc). So in this case `sendNative` and `unwrapNative` will be disabled. If funds are deposited anyway refund address can still refund

sammy1991106 (approved):

milapsheth (approved):

PR #20, opened 2021-09-20
feat: Remove SafeMath for solc 0.8.x

deluca-mike:
Solidity 0.8.x does built-in math checks.

Issue #19, opened 2021-09-20
Controller Service


wilcollins:
https://www.notion.so/Controller-Service-9e11af5d54884c98a937269fa8434c1f

PR #160, opened 2022-11-16
feat(GMP): adding GMP Express Service

re1ro:
* [x] `GMPExpressService`: new service that calls ExpressExecutable contracts
* [x] Utils: using dependencies from latest `axelar-gmp-sdk-solidity`
* [x] Fixed existing tests
* [ ] TODO: more tests for `GMPExpressService` and `GasService`
* [x] Dependency: https://github.com/axelarnetwork/axelar-gmp-sdk-solidity/pull/19
* [x] Dependency: https://github.com/axelarnetwork/axelar-gmp-sdk-solidity/pull/20


contracts/forecall-service/AxelarForecallService.sol
```diff
@@ -0,0 +1,126 @@
...
+                if (
+                    gateway.isContractCallAndMintApproved(
+                        commandId,
+                        sourceChain,
+                        sourceAddress,
+                        contractAddress,
+                        payloadHash,
+                        tokenSymbolA,
+                        amount
+                    )
```
Foivos, line 78: You can omit this to save gas since executeWithToken will revert anyway.

contracts/forecall-service/AxelarForecallService.sol
```diff
@@ -0,0 +1,126 @@
...
+    ) external onlyOperator {
+        if (contractAddress == address(0)) revert InvalidContractAddress();
+
+        if (commandId == bytes32(0)) {
+            IAxelarForecallable(contractAddress).forecall(sourceChain, sourceAddress, payload);
+        } else {
+            if (gateway.isCommandExecuted(commandId)) {
+                bytes32 payloadHash = keccak256(payload);
+
+                if (gateway.isContractCallApproved(commandId, sourceChain, sourceAddress, contractAddress, payloadHash)) {
```
Foivos, line 41: You can omit this to save gas since execute will revert anyway.

contracts/forecall-service/AxelarForecallService.sol
```diff
@@ -0,0 +1,126 @@
...
+import { IAxelarGateway } from '../interfaces/IAxelarGateway.sol';
+import { IAxelarForecallService } from '../interfaces/IAxelarForecallService.sol';
+import { IERC20 } from '../interfaces/IERC20.sol';
+
+// This should be owned by the microservice that is paying for gas.
+contract AxelarForecallService is Upgradable, IAxelarForecallService {
+    IAxelarGateway public immutable gateway;
+    address public immutable forecallOperator;
+
+    constructor(address gateway_, address forecallOperator_) {
```
eguajardo, line 15: shouldn't this be moved to a "setup" function to work with the proxy init?

contracts/forecall-service/AxelarForecallService.sol
milapsheth: I think it's preferable to not rely on the app contract to revert

contracts/forecall-service/AxelarForecallService.sol
```diff
@@ -0,0 +1,126 @@
...
+        if (contractAddress == address(0)) revert InvalidContractAddress();
+
+        if (commandId == bytes32(0)) {
+            IAxelarForecallable(contractAddress).forecall(sourceChain, sourceAddress, payload);
+        } else {
+            if (gateway.isCommandExecuted(commandId)) {
+                bytes32 payloadHash = keccak256(payload);
+
+                if (gateway.isContractCallApproved(commandId, sourceChain, sourceAddress, contractAddress, payloadHash)) {
+                    IAxelarForecallable(contractAddress).execute(commandId, sourceChain, sourceAddress, payload);
```
milapsheth, line 42: not sure if we should execute. this will conflict a bit with the executor service if tries to execute at the same time (without knowledge of the forecaller)

contracts/forecall-service/AxelarForecallService.sol
re1ro: Those are immutable variables that can be initialized only from constructor. They don't cost us storage as they become part of the bytecode

contracts/forecall-service/AxelarForecallService.sol
```diff
@@ -0,0 +1,126 @@
...
+        if (receiver == address(0)) revert InvalidAddress();
+
+        if (token == address(0)) {
+            receiver.transfer(amount);
+        } else {
+            _safeTransfer(token, receiver, amount);
+        }
+    }
+
+    function _safeTransfer(
```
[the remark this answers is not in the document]
re1ro: * [ ] TODO

contracts/forecall-service/AxelarForecallService.sol
re1ro: Executor service will not run as gas for paid for the Express Execute service

PR #161, opened 2022-12-03
fix(Gateway): exit command execution if no gas left

re1ro:
* [x] Exit from command execution loop if no gas left

### DIFF
--- a/contracts/AxelarGateway.sol
+++ b/contracts/AxelarGateway.sol
@@ -324,13 +324,25 @@ contract AxelarGateway is IAxelarGateway, AdminMultisigBase {
                 continue; /* Ignore if unknown command received */
             }
 
+            uint256 availableGas = gasleft();
+            if (availableGas < 100000) return;
+
             // Prevent a re-entrancy from executing this command before it can be marked as successful.
             _setCommandExecuted(commandId, true);
             // solhint-disable-next-line avoid-low-level-calls
-            (bool success, ) = address(this).call(abi.encodeWithSelector(commandSelector, params[i], commandId));
+            (bool success, ) = address(this).call{ gas: availableGas - 10000 }(
+                abi.encodeWithSelector(commandSelector, params[i], commandId)
+            );
 
             if (success) emit Executed(commandId);
             else _setCommandExecuted(commandId, false);
+
+            uint256 remainingGas = gasleft();
+
+            // Exit if spent too much gas
+            if ((commandHash != SELECTOR_TRANSFER_OPERATORSHIP && availableGas - remainingGas > 1000000)) return;
+            // Exit if not enough left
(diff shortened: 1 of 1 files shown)


contracts/AxelarGateway.sol
```diff
@@ -324,13 +324,25 @@ contract AxelarGateway is IAxelarGateway, AdminMultisigBase {
...
 
             if (success) emit Executed(commandId);
             else _setCommandExecuted(commandId, false);
+
+            uint256 remainingGas = gasleft();
+
+            // Exit if spent too much gas
+            if ((commandHash != SELECTOR_TRANSFER_OPERATORSHIP && availableGas - remainingGas > 1000000)) return;
+            // Exit if not enough left
+            if (remainingGas < 10000) return;
```
milapsheth, line 345:
```suggestion
            if (remainingGas < 100000) return;
```
let's keep 100k room just to be safe?

contracts/AxelarGateway.sol
```diff
@@ -324,13 +324,25 @@ contract AxelarGateway is IAxelarGateway, AdminMultisigBase {
...
+                abi.encodeWithSelector(commandSelector, params[i], commandId)
+            );
 
             if (success) emit Executed(commandId);
             else _setCommandExecuted(commandId, false);
+
+            uint256 remainingGas = gasleft();
+
+            // Exit if spent too much gas
+            if ((commandHash != SELECTOR_TRANSFER_OPERATORSHIP && availableGas - remainingGas > 1000000)) return;
```
milapsheth, line 343: given that refunds only occur at the end of tx, we've seen examples where burns cost more than >1M. I think we shouldn't add this check since even if there is such a cmd, we kept enough gas budget to exit.

contracts/AxelarGateway.sol
```diff
@@ -324,13 +324,25 @@ contract AxelarGateway is IAxelarGateway, AdminMultisigBase {
...
             }
 
+            uint256 availableGas = gasleft();
+            if (availableGas < 100000) return;
+
             // Prevent a re-entrancy from executing this command before it can be marked as successful.
             _setCommandExecuted(commandId, true);
             // solhint-disable-next-line avoid-low-level-calls
-            (bool success, ) = address(this).call(abi.encodeWithSelector(commandSelector, params[i], commandId));
+            (bool success, ) = address(this).call{ gas: availableGas - 10000 }(
```
milapsheth, line 333:
```suggestion
            (bool success, ) = address(this).call{ gas: availableGas - 100000 }(
```
I prefer keeping a higher budget to be safe. doesn't require us to pay more gas, just decreases the max batching possibility by a bit.

re1ro: Yeah that makes sense. I will close this PR in favor of the actual fix

PR #174, opened 2023-03-22
feat(GMP-SDK): safe transfer

re1ro:
* [x] Safe transfer for ERC20 and native currency
* [x] Using `@axelar-network/axelar-gmp-sdk-solidity@3.0.0` as a dependency for safe transfer
* [x] Using upgrade scripts from the SDK


sammy1991106 (approved):

contracts/deposit-service/AxelarDepositService.sol
```diff
@@ -255,15 +260,12 @@ contract AxelarDepositService is Upgradable, DepositServiceBase, IAxelarDepositS
...
         if (receiver == address(0)) revert InvalidAddress();
+        if (amount == 0) revert InvalidAmount();
 
         if (token == address(0)) {
-            if (amount == 0) revert InvalidAmount();
-
-            (bool sent, ) = receiver.call{ value: amount }('');
-
-            if (!sent) revert NativeTransferFailed();
+            receiver.safeNativeTransfer(amount);
```
milapsheth, line 266: might need to allow passing in more gas on some chains, where it naturally costs more than 2300. We can add that ability if needed

contracts/deposit-service/DepositServiceBase.sol
```diff
@@ -2,13 +2,15 @@
...
+import { SafeTokenTransfer } from '@axelar-network/axelar-gmp-sdk-solidity/contracts/utils/SafeTransfer.sol';
 import { IAxelarGateway } from '../interfaces/IAxelarGateway.sol';
 import { IDepositServiceBase } from '../interfaces/IDepositServiceBase.sol';
-import { IERC20 } from '../interfaces/IERC20.sol';
 import { Bytes32ToString, StringToBytes32 } from '../util/BytesStringUtil.sol';
 
 // This should be owned by the microservice that is paying for gas.
-contract DepositServiceBase is IDepositServiceBase {
+abstract contract DepositServiceBase is IDepositServiceBase {
+    using SafeTokenTransfer for address;
```
milapsheth, line 12:
```suggestion
    using SafeTokenTransfer for IERC20;
```
typo? also doesn't look like it's being used here

scripts/deploy-upgradable.js
```diff
@@ -2,18 +2,24 @@
...
 const {
+    Contract,
     Wallet,
     getDefaultProvider,
     utils: { isAddress },
 } = require('ethers');
-const { deployUpgradable, upgradeUpgradable, getProxy, predictProxyAddress } = require('./upgradable');
 const readlineSync = require('readline-sync');
 const { outputJsonSync } = require('fs-extra');
+const { deployUpgradable, upgradeUpgradable } = require('@axelar-network/axelar-gmp-sdk-solidity');
```
milapsheth, line 12: The `upgradable.js` in gmp sdk repo should be updated to with the changes present in this one before we delete it. For e.g. `predictProxyAddress` is being removed here. Can you do a diff and port over the changes there?

re1ro: They those methods are already present in `gmp-sdk`

contracts/deposit-service/DepositServiceBase.sol
re1ro: Yup removed

scripts/deploy-upgradable.js
re1ro: Fixed by using `predictContractConstant` directly from `SDK`

milapsheth (approved):

PR #176, opened 2023-03-27
feat(GMP-SDK): using Upgradable

re1ro:
* [x] `Upgradable`: importing from GMP-SDK


sammy1991106 (approved):

contracts/deposit-service/AxelarDepositService.sol
```diff
@@ -2,12 +2,12 @@
 
 pragma solidity 0.8.9;
 
-import { SafeTokenTransfer, SafeNativeTransfer } from '@axelar-network/axelar-gmp-sdk-solidity/contracts/utils/SafeTransfer.sol';
 import { IERC20 } from '@axelar-network/axelar-gmp-sdk-solidity/contracts/interfaces/IERC20.sol';
+import { Upgradable } from '@axelar-network/axelar-gmp-sdk-solidity/contracts/upgradable/Upgradable.sol';
```
milapsheth, line 6: Don't want to switch to the new upgradable in this release, since we have a few other things to focus on. Let's merge it a bit later. I wanna make sure there are no unexpected issues when switching and have folks be comfortable with the new ownership transfer method.

contracts/gas-service/AxelarGasServiceProxy.sol
```diff
@@ -3,7 +3,6 @@
 pragma solidity 0.8.9;
 
 import { Proxy } from '../util/Proxy.sol';
-import { IUpgradable } from '../interfaces/IUpgradable.sol';
```
milapsheth, line 6: does this affect the bytecode?

PR #180, opened 2023-04-03
test(ServiceProxy): should preserve Proxy bytecode

re1ro:
- [x] `DepositService`, `GasService`: should preserve Proxy bytecode to get the same address across chains

### DIFF
--- a/test/AxelarDepositService.js
+++ b/test/AxelarDepositService.js
@@ -355,6 +355,12 @@ describe('AxelarDepositService', () => {
             ).to.changeEtherBalance(userWallet, amount);
         });
 
+        it('should have the same receiver bytecode', async () => {
+            await expect(keccak256(DepositReceiver.bytecode)).to.be.equal(
+              '0xc0fd88839756e97f51ab0395ce8e6164a5f924bd73a3342204340a14ad306fe1',
+            );
+        });
+
         it('should refund have the same proxy bytecode', async () => {
             const proxyBytecode = DepositServiceProxy.bytecode;
             const proxyBytecodeHash = keccak256(proxyBytecode);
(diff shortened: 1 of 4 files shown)


milapsheth (approved):

test/AxelarDepositService.js
```diff
@@ -354,5 +357,30 @@ describe('AxelarDepositService', () => {
...
+            await expect(keccak256(DepositReceiver.bytecode)).to.be.equal(
+              expected,
+            );
+        });
+
+        it('should have the same proxy bytecode', async () => {
+            const proxyBytecode = DepositServiceProxy.bytecode;
+            const proxyBytecodeHash = keccak256(proxyBytecode);
+
+            const expected = {
```
sammy1991106, line 377: These proxy bytecode doesn't look the same while the name of the test case suggests that they should be?

milapsheth: Most of the contracts are compiled differently when targeting a different EVM version. Maybe @re1ro can mention if there's a specific opcode where they differ just to get a better idea. But not really expecting the same one to be generated.

test/AxelarGateway.js
```diff
@@ -728,6 +730,42 @@ describe('AxelarGateway', () => {
...
+                istanbul: '0xc68014e297eb42dbde383254ef3129d59528159e6c51b4f9a38f995be1dd451f',
+                berlin: '0xd3a39792ca8d1ce8e5318135ca29d8a7f0b800837726997b132ebc04f88cf9aa',
+                london: '0x0698929742de660596af20d09d04eb91bfe532ef5e2927858e4c4952034967a5',
+            }[EVM_VERSION]
+
+            await expect(keccak256(tokenDeployerFactory.bytecode)).to.be.equal(
+              expectedDeployer,
+            );
+        });
     });
```
milapsheth, line 769: Can you add the AxelarGatewayProxy and gateway implementation bytecode too? While we don't need it for deriving the same address, we do want to be alerted about any unintentional changes

test/AxelarDepositService.js
deanamiel: That confused me a bit as well, I think changing the test name would help

milapsheth (approved):

deanamiel (approved):

sammy1991106 (approved):

PR #185, opened 2023-04-11
feat: switch to hardhat toolbox

milapsheth:
- Switch to hardhat toolbox
- Fix GMP tests to use hardhat
- Remove waffle dependency
- Update deps to recent versions


test/AxelarGateway.js
```diff
@@ -352,10 +352,7 @@ describe('AxelarGateway', () => {
                 threshold,
                 operators.slice(0, threshold),
             );
-            await expect(gateway.execute(secondInput))
-                .to.not.emit(gateway, 'TokenDeployed')
-                .and.to.emit(gateway, 'Executed')
-                .withArgs(secondCommandID);
+            await expect(gateway.execute(secondInput)).to.not.emit(gateway, 'Executed').to.emit(gateway, 'TokenDeployed'); // TODO: fix
```
milapsheth, line 355: Found a bug with our chai usage, this should actually fail because the event shouldn't be emitted. Can't figure out the right usage here. `.and` doesn't do anything, need a way to combine multiple tests.

re1ro (approved):

PR #190, opened 2023-04-26
fix: deploy deposit service

deanamiel:
Quick fix to deploy deposit service until GMP express branch gets merged

### DIFF
--- a/scripts/deploy-upgradable.js
+++ b/scripts/deploy-upgradable.js
@@ -1,6 +1,6 @@
 'use strict';
 require('dotenv').config();
-const _ = require('lodash/fp');
+const { get, getOr, isEmpty } = require('lodash/fp');
 const {
     Contract,
     Wallet,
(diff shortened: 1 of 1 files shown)


scripts/deploy-upgradable.js
```diff
@@ -103,7 +103,7 @@ async function deploy(env, chains, wallet, artifactPath, contractName, deployTo)
                 implementationJson,
                 args,
                 getUpgradeArgs(contractName, chain),
-                chain.gasOptions,
+                _.get('gasOptions.gasLimit', chain),
```
milapsheth, line 106: Ah, I see. Should refactor this in future to take all gas options (might need to override gas price, gas tip etc.).

milapsheth (approved):

scripts/deploy-upgradable.js
sammy1991106, line 106: Change the import to be `const { get } = require('lodash/fp');` so that here it becomes `get('gasOptions.gasLimit', chain),`

Foivos (approved):

scripts/deploy-upgradable.js
sammy1991106: yes

deanamiel: sounds good, just updated

sammy1991106 (approved):

PR #193, opened 2023-05-10
feat: added a separate refund method that takes txHash/logIndex and emit a Refunded event

sammy1991106:
https://axelarworkspace.slack.com/archives/C02T1BNNS9Y/p1683245157722149

### DIFF
--- a/contracts/interfaces/IAxelarGasService.sol
+++ b/contracts/interfaces/IAxelarGasService.sol
@@ -84,6 +84,8 @@ interface IAxelarGasService is IUpgradable {
 
     event NativeExpressGasAdded(bytes32 indexed txHash, uint256 indexed logIndex, uint256 gasFeeAmount, address refundAddress);
 
+    event Refunded(bytes32 indexed txHash, uint256 indexed logIndex, address payable receiver, address token, uint256 amount);
+
     // This is called on the source chain before calling the gateway to execute a remote contract.
     function payGasForContractCall(
         address sender,
--- a/test/gmp/AxelarGasService.js
+++ b/test/gmp/AxelarGasService.js
@@ -3,7 +3,8 @@
 const chai = require('chai');
 const { ethers, config } = require('hardhat');
 const {
-    utils: { defaultAbiCoder, keccak256, parseEther },
+    utils: { defaultAbiCoder, keccak256, parseEther, id },
+    constants: { AddressZero, HashZero },
 } = ethers;
 const { expect } = chai;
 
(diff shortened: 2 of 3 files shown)


contracts/interfaces/IAxelarGasService.sol
```diff
@@ -84,6 +84,8 @@ interface IAxelarGasService is IUpgradable {
 
     event NativeExpressGasAdded(bytes32 indexed txHash, uint256 indexed logIndex, uint256 gasFeeAmount, address refundAddress);
 
+    event Refunded(bytes32 indexed txHash, uint256 indexed logIndex, address payable receiver, address token, uint256 amount);
```
milapsheth, line 87:
```suggestion
    event Refunded(bytes32 indexed txHash, uint256 indexed logIndex, address payable indexed receiver, address token, uint256 amount);
```
Keep the receiver as indexed too, so we can query refunds for specific users easily?

test/gmp/AxelarGasService.js
```diff
@@ -299,37 +300,171 @@ describe('AxelarGasService', () => {
...
+                gasService
+                    .connect(userWallet)
+                    .functions['refund(address,address,uint256)'](userWallet.address, testToken.address, gasFeeAmount),
+            ).to.be.reverted;
 
-            await expect(await gasService.connect(ownerWallet).refund(userWallet.address, testToken.address, gasFeeAmount))
+            await expect(
+                gasService
+                    .connect(ownerWallet)
+                    .functions['refund(address,address,uint256)'](userWallet.address, testToken.address, gasFeeAmount),
```
milapsheth, line 327:
```suggestion
                    .refund(userWallet.address, testToken.address, gasFeeAmount),
```
Why not just call it directly everywhere? It'll be more readable

sammy1991106: nope js doesn't support function overloading, and hardhat isn't smart enough to work with it.

contracts/interfaces/IAxelarGasService.sol
sammy1991106: Kane didn't say it's needed. Adding it would make it not able to search by `txHash` and `logIndex` pair alone.

milapsheth: Ah, that's a good point

test/gmp/AxelarGasService.js
sammy1991106: I think the contract itself being readable is more important than the js testing. I'd slightly prefer a different name if we plan to keep both functions. However, in this case of one being deprecated and potential removed in the near future, I'd say this ugliness in the js code is a temporary thing anyway.

sammy1991106: `refundExcessGas` makes sense if the function does the calculation. In our case, it's all up to the collector so the "excess" part is confusing.

milapsheth: Yeah, the concern is lower due to that. I do think the name I mentioned above is more readable than just refund, but not a strong opinion.

milapsheth (approved):

contracts/interfaces/IAxelarGasService.sol
kane-armstrong, line 87: This differs from the precedent established by the other events - those suggest an approach like this:

```
event NativeGasRefunded(bytes32 indexed txHash, uint256 indexed logIndex, address payable receiver, uint256 amount);

event GasRefunded(bytes32 indexed txHash, uint256 indexed logIndex, address payable receiver, address token, uint256 amount); 
```

A single event is pragmatic though, since it's not difficult to check for `address(0)` to figure out what token was refunded. No problem from a consumer POV, just thought I'd point it out

kane-armstrong (approved):

re1ro (approved):

contracts/interfaces/IAxelarGasService.sol
milapsheth: Since the method used is the same for either case, I think we should use the same event

PR #194, opened 2023-05-27
feature(Governance): initial governance executable implementation

re1ro:
* [x] `AxelarGatewayGovernance`: based on `TimeLock` 
* [x] `AxelarServiceGovernance`: based on `TimeLock` + `MultisigBase`
* [x] `AxelarMultisigMintLimiter`: based on `MultisigBase`


contracts/AxelarGateway.sol
```diff
@@ -50,13 +50,21 @@ contract AxelarGateway is IAxelarGateway, AdminMultisigBase {
     // solhint-disable-next-line var-name-mixedcase
     address internal immutable AUTH_MODULE;
     // solhint-disable-next-line var-name-mixedcase
+    address internal immutable GOVERNANCE_MODULE;
```
milapsheth, line 53: So to allow a change of governance modules, there needs to be an upgrade to a new instance of the implementation with a difference governance module account. Not as explicit but it's more gas efficient so works i think

contracts/governance/AdminMultisigBase.sol
```diff
@@ -0,0 +1,219 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity 0.8.9;
+
+import { IAdminMultisig } from '../interfaces/IAdminMultisig.sol';
+import { EternalStorage } from '../EternalStorage.sol';
+
+contract AdminMultisigBase is EternalStorage, IAdminMultisig {
```
milapsheth, line 8: Need to think of a different name, since these addresses are more like co-signers rather than governance admins

contracts/governance/AxelarGovernance.sol
```diff
@@ -0,0 +1,87 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity 0.8.9;
+
+import { Upgradable } from '@axelar-network/axelar-gmp-sdk-solidity/contracts/upgradable/Upgradable.sol';
+import { IAxelarGateway } from '@axelar-network/axelar-gmp-sdk-solidity/contracts/interfaces/IAxelarGateway.sol';
+import { IAxelarGovernance } from '../interfaces/IAxelarGovernance.sol';
+import { AdminMultisigBase } from './AdminMultisigBase.sol';
+
+contract AxelarGovernance is Upgradable, AdminMultisigBase, IAxelarGovernance {
```
milapsheth, line 10: Shouldn't be upgradable to not allow admins to bypass the governance

contracts/governance/AxelarGovernance.sol
```diff
@@ -0,0 +1,87 @@
...
+        } else if (command == Command.SetMintLimits) {
+            if (methodSelector != IAxelarGateway.setTokenMintLimits.selector) revert InvalidCommand();
+        } else {
+            revert InvalidCommand();
+        }
+
+        (bool success, ) = targetContract.call(callData);
+
+        if (!success) {
+            revert ExecutionFailed();
```
milapsheth, line 58: we should add a `SafeCall` helper

contracts/governance/AxelarGovernance.sol
milapsheth, line 10: And we need a separate contract that doesn't check for governance for the separate operator role. Could technically reuse the same contract and governance address to `0`, but I'd prefer to be more explicit

milapsheth, line 10: Inherit from AxelarExecutable and override `_execute` instead?

contracts/AxelarGateway.sol
```diff
@@ -50,13 +50,21 @@ contract AxelarGateway is IAxelarGateway, AdminMultisigBase {
...
+    // solhint-disable-next-line var-name-mixedcase
     address internal immutable TOKEN_DEPLOYER_IMPLEMENTATION;
 
-    constructor(address authModule_, address tokenDeployerImplementation_) {
```
milapsheth, line 55: Add an ability to let the governance module pause the gateway, i.e execute can only be called when not paused, to have an emergency mechanism for the pure GMP calls

contracts/interfaces/ICosignedBase.sol
```diff
@@ -0,0 +1,21 @@
...
+
+    /***********\
+    |* Getters *|
+    \***********/
+
+    function cosignerEpoch() external view returns (uint256);
+
+    function cosignerThreshold(uint256 epoch) external view returns (uint256);
+
+    function cosignerAccounts(uint256 epoch) external view returns (address[] memory);
```
milapsheth, line 20:
```suggestion
    function threshold() external view returns (uint256);

    function signers() external view returns (address[] memory);
```

contracts/auth/CosignedBase.sol
```diff
@@ -0,0 +1,104 @@
...
+
+        for (uint256 i; i < length; ++i) {
+            address account = newAccounts[i];
+
+            // Check that the account wasn't already set as a cosigner for this epoch.
+            if (newSigners.isCosigner[account]) revert DuplicateCosigner(account);
+            if (account == address(0)) revert InvalidCosigners();
+
+            newSigners.isCosigner[account] = true;
+        }
```
milapsheth, line 102: add an event for the new signers + threshold

contracts/auth/CosignedBase.sol
```diff
@@ -0,0 +1,104 @@
...
+
+        // Determine the new vote count.
+        uint256 voteCount = voting.voteCount + 1;
+
+        // Do not proceed with operation execution if insufficient votes.
+        if (voteCount < signers.threshold) {
+            // Save updated vote count.
+            voting.voteCount = voteCount;
+            return;
+        } else if (voteCount > 1) {
```
milapsheth, line 48: t = 1 optimization isn't needed anymore

contracts/auth/CosignedBase.sol
```diff
@@ -0,0 +1,104 @@
...
+            // Clear vote count and voted booleans.
+            voting.voteCount = 0;
+
+            uint256 count = signers.accounts.length;
+
+            for (uint256 i; i < count; ++i) {
+                voting.hasVoted[signers.accounts[i]] = false;
+            }
+        }
+
```
milapsheth, line 58: emit a `Voted(topic)` event perhaps?

contracts/AxelarGateway.sol
```diff
@@ -197,34 +237,19 @@ contract AxelarGateway is IAxelarGateway, AdminMultisigBase {
...
-        uint256 adminCount = _getAdminCount(epoch);
-        results = new address[](adminCount);
+    /************************\
+    |* Governance Functions *|
+    \************************/
 
-        for (uint256 i; i < adminCount; ++i) {
-            results[i] = _getAdmin(epoch, i);
-        }
+    function pauseGateway(bool isPaused) external override onlyGovernance {
```
sammy1991106, line 244:
```suggestion
    function pause(bool isPaused) external override onlyGovernance {
```

contracts/AxelarGateway.sol
```diff
@@ -281,7 +298,7 @@ contract AxelarGateway is IAxelarGateway, AdminMultisigBase {
         }
     }
 
-    function execute(bytes calldata input) external override {
+    function execute(bytes calldata input) external override whenNotPaused {
```
sammy1991106, line 301: This would prevent the upgrade from happening though.

contracts/governance/AxelarGatewayGovernance.sol
```diff
@@ -0,0 +1,87 @@
...
+pragma solidity 0.8.9;
+
+import { AxelarExecutable } from '@axelar-network/axelar-gmp-sdk-solidity/contracts/executable/AxelarExecutable.sol';
+import { IAxelarGovernance } from '../interfaces/IAxelarGovernance.sol';
+import { TimeLock } from '../util/TimeLock.sol';
+
+contract AxelarGatewayGovernance is AxelarExecutable, TimeLock, IAxelarGovernance {
+    enum Command {
+        ScheduleProposal,
+        CancelProposal
```
sammy1991106, line 12:
```suggestion
        ScheduleUpgrade,
        CancelUpgrade
```

contracts/governance/AxelarServiceGovernance.sol
```diff
@@ -0,0 +1,23 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity 0.8.9;
+
+import { AxelarGatewayGovernance } from './AxelarGatewayGovernance.sol';
+import { MultisigBase } from '../auth/MultisigBase.sol';
+
+contract AxelarServiceGovernance is MultisigBase, AxelarGatewayGovernance {
```
sammy1991106, line 8: what's `AxelarServiceGovernance` v.s. `AxelarGatewayGovernance`?

contracts/AxelarGateway.sol
```diff
@@ -30,6 +30,7 @@ contract AxelarGateway is IAxelarGateway, AdminMultisigBase {
 
     /// @dev Storage slot with the address of the current implementation. `keccak256('eip1967.proxy.implementation') - 1`.
     bytes32 internal constant KEY_IMPLEMENTATION = bytes32(0x360894a13ba1a3210667c828492db98dca3e2076cc3735a920a3ca505d382bbc);
+    bytes32 internal constant KEY_PAUSED = keccak256('gateway-paused');
```
milapsheth, line 33: We decided to not add pause functionality

contracts/auth/MultisigBase.sol
```diff
@@ -0,0 +1,111 @@
...
+    struct Signers {
+        address[] accounts;
+        uint256 threshold;
+        mapping(address => bool) isSigner;
+        // bytes32 is for vote topic hash
+        mapping(bytes32 => Voting) voting;
+    }
+
+    uint256 public signerEpoch;
+    mapping(uint256 => Signers) public signersPerEpoch;
```
milapsheth, line 22: Epoch is only needed for domain separation for vote tracking. So, I think we should remove accounts, threshold, isSigner into a separate struct that can be overwritten instead of being epoch separated. This should save some gas too I imagine

milapsheth: A benefit of this is that queries for the current signer set don't require epoch to be passed in.

Gas savings matter less, since this code isn't triggered too often

contracts/AxelarGateway.sol
```diff
@@ -50,13 +51,26 @@ contract AxelarGateway is IAxelarGateway, AdminMultisigBase {
...
-    constructor(address authModule_, address tokenDeployerImplementation_) {
+    constructor(
+        address authModule_,
+        address governanceModule_,
+        address mintLimitModule_,
+        address tokenDeployerImplementation_
+    ) {
         if (authModule_.code.length == 0) revert InvalidAuthModule();
+        if (governanceModule_.code.length == 0) revert InvalidGovernanceModule();
+        if (mintLimitModule_.code.length == 0) revert InvalidGovernanceModule();
```
milapsheth, line 68:
```suggestion
        if (mintLimitModule_.code.length == 0) revert InvalidMintLimitModule();
```
Or use a more generic error for all of them such as `InvalidContractAddress`/`NotAContract`

contracts/governance/AxelarMultisigMintLimiter.sol
```diff
@@ -0,0 +1,19 @@
...
+
+import { IMultisigExecute } from '../interfaces/IMultisigExecute.sol';
+import { MultisigBase } from '../auth/MultisigBase.sol';
+
+contract AxelarMultisigMintLimiter is MultisigBase, IMultisigExecute {
+    constructor(address[] memory accounts, uint256 threshold) {
+        _rotateSigners(accounts, threshold);
+    }
+
+    function execute(address target, bytes calldata callData) external onlySigners {
```
milapsheth, line 13: add ether value that should be forwarded to the target as an arg. Makes it more generic, allows withdrawing ether for e.g. etc
```suggestion
    function execute(address target, bytes calldata args, uint value) external onlySigners {
```

contracts/util/TimeLock.sol
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    function _scheduleTimeLock(bytes32 hash, uint256 eta) internal {
+        if (hash == 0) revert InvalidTimeLockHash();
+
+        uint256 minimumEta = block.timestamp + MINIMUM_TIME_LOCK_DELAY;
+
+        if (eta < minimumEta) eta = minimumEta;
+
+        if (timeLockQueue[hash] != 0) revert TimeLockAlreadyScheduled();
```
milapsheth, line 23: check this first

contracts/util/TimeLock.sol
```diff
@@ -0,0 +1,43 @@
...
+
+        timeLockQueue[hash] = eta;
```
milapsheth, line 25: add an event so implementations benefit from a common event. implementations can still more readable events at a higher level if they want

contracts/governance/AxelarGatewayGovernance.sol
milapsheth: not just for upgrades. governance can set rate limits too for e.g., so we should keep it generic

contracts/interfaces/IAxelarGovernance.sol
```diff
@@ -0,0 +1,20 @@
...
+    error InvalidTargetContract();
+    error InvalidCallData();
+    error ExecutionFailed();
+    error TokenNotSupported();
+
+    event ProposalScheduled(bytes32 indexed proposalHash, address indexed targetContract, bytes callData, uint256 eta);
+    event ProposalCancelled(bytes32 indexed proposalHash);
+    event ProposalExecuted(bytes32 indexed proposalHash);
+
+    function executeProposal(address targetContract, bytes calldata callData) external;
```
milapsheth, line 19:
```suggestion
    function executeProposal(address target, bytes calldata callData) external;
```
target might be an account, where we're transferring ether

contracts/auth/MultisigBase.sol
re1ro: You can't overwrite struct in solidity. That's why it needs to be epoch separated

re1ro:
Closed in favor of #196 #197 #198 #199 #203

PR #195, opened 2023-06-07
feat: operators contract as per ticket axe680

deanamiel:
Draft for the contract described in ticket AXE-680. It is a pretty straightforward implementation of having multiple operators any of whom can call any arbitrary contract. There is also onlyOwner functionality to add and remove operators.

### DIFF
--- a/contracts/interfaces/IOperators.sol
+++ b/contracts/interfaces/IOperators.sol
@@ -6,7 +6,7 @@ interface IOperators {
     error NotOperator();
     error InvalidOperator();
     error OperatorAlreadyAdded();
-    error OperatorAlreadyRemoved();
+    error NotAnOperator();
     error ExecutionFailed();
 
     event OperatorAdded(address indexed operator);
--- a/contracts/interfaces/IOperators.sol
+++ b/contracts/interfaces/IOperators.sol
@@ -9,14 +9,14 @@ interface IOperators {
     error OperatorAlreadyRemoved();
     error ExecutionFailed();
 
-    event OperatorAdded(address indexed newOperator);
-    event OperatorRemoved(address indexed oldOperator);
+    event OperatorAdded(address indexed operator);
+    event OperatorRemoved(address indexed operator);
 
-    function isOperator(address operator) external view returns (bool);
+    function isOperator(address account) external view returns (bool);
 
-    function addOperator(address newOperator) external;
+    function addOperator(address operator) external;
 
-    function removeOperator(address oldOperator) external;
+    function removeOperator(address operator) external;
 
-    function execute(address target, bytes calldata callData) external;
+    function execute(address target, bytes calldata callData) external returns (bytes memory);
 }
--- a/contracts/interfaces/IOperators.sol
+++ b/contracts/interfaces/IOperators.sol
@@ -1,0 +1,22 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.9;
+
+interface IOperators {
+    error NotOperator();
+    error InvalidOperator();
+    error OperatorAlreadyAdded();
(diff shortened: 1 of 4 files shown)


contracts/util/Operators.sol
```diff
@@ -0,0 +1,46 @@
...
+    modifier onlyOperator() {
+        if (!operators[msg.sender]) revert NotOperator();
+        _;
+    }
+
+    function isOperator(address _operator) external view returns (bool) {
+        return operators[_operator];
+    }
+
+    function addOperator(address _newOperator) external onlyOwner {
```
milapsheth, line 22:
```suggestion
    function addOperator(address operator) external onlyOwner {
```

contracts/util/Operators.sol
```diff
@@ -0,0 +1,46 @@
...
+        if (_newOperator == address(0)) revert InvalidOperator();
+        if (operators[_newOperator]) revert OperatorAlreadyAdded();
+
+        operators[_newOperator] = true;
+
+        emit OperatorAdded(_newOperator);
+    }
+
+    function removeOperator(address _oldOperator) external onlyOwner {
```
milapsheth, line 31:
```suggestion
    function removeOperator(address operator) external onlyOwner {
```

contracts/util/Operators.sol
```diff
@@ -0,0 +1,46 @@
...
+        if (_oldOperator == address(0)) revert InvalidOperator();
+        if (!operators[_oldOperator]) revert OperatorAlreadyRemoved();
+
+        operators[_oldOperator] = false;
+
+        emit OperatorRemoved(_oldOperator);
+    }
+
+    function execute(address target, bytes calldata callData) external onlyOperator {
```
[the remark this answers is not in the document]
deanamiel: Sure, I'll store the return data from the .call() and return it

contracts/util/Operators.sol
```diff
@@ -15,32 +15,34 @@ contract Operators is Ownable, IOperators {
...
-        emit OperatorAdded(_newOperator);
+        emit OperatorAdded(operator);
     }
 
-    function removeOperator(address _oldOperator) external onlyOwner {
-        if (_oldOperator == address(0)) revert InvalidOperator();
-        if (!operators[_oldOperator]) revert OperatorAlreadyRemoved();
+    function removeOperator(address operator) external onlyOwner {
+        if (operator == address(0)) revert InvalidOperator();
+        if (!operators[operator]) revert OperatorAlreadyRemoved();
```
milapsheth, line 33: Looks good, can you add tests?

contracts/interfaces/IOperators.sol
```diff
@@ -9,14 +9,14 @@ interface IOperators {
     error OperatorAlreadyRemoved();
```
milapsheth, line 9: Maybe `NotAnOperator` error instead (since it could have never been an operator to begin with).

contracts/util/Operators.sol
```diff
@@ -15,32 +15,34 @@ contract Operators is Ownable, IOperators {
...
+        operators[operator] = false;
 
-        emit OperatorRemoved(_oldOperator);
+        emit OperatorRemoved(operator);
     }
 
-    function execute(address target, bytes calldata callData) external onlyOperator {
-        (bool success, ) = target.call(callData);
+    function execute(address target, bytes calldata callData) external onlyOperator returns (bytes memory) {
+        (bool success, bytes memory data) = target.call(callData);
```
milapsheth, line 41: We should add a SafeCall in gmp sdk to abstract this. There's something for tokens already, could generalize that

milapsheth:
moved to gmp-sdk repo

PR #196, opened 2023-06-20
feat(Governance): new module based on TimeLock

re1ro:
* [x] `Governance`: new module based on TimeLock
* [x] Depends on https://github.com/axelarnetwork/axelar-gmp-sdk-solidity/pull/40

### DIFF
--- a/contracts/governance/AxelarGovernance.sol
+++ b/contracts/governance/AxelarGovernance.sol
@@ -1,0 +1,87 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity 0.8.9;
+
+import { AxelarExecutable } from '@axelar-network/axelar-gmp-sdk-solidity/contracts/executable/AxelarExecutable.sol';
+import { TimeLock } from '@axelar-network/axelar-gmp-sdk-solidity/contracts/utils/TimeLock.sol';
+import { IAxelarGovernance } from '../interfaces/IAxelarGovernance.sol';
+
+contract AxelarGovernance is AxelarExecutable, TimeLock, IAxelarGovernance {
+    enum Command {
+        ScheduleProposal,
+        CancelProposal
+    }
+
+    bytes32 public immutable governanceChainHash;
+    bytes32 public immutable governanceAddressHash;
+
+    // solhint-disable-next-line no-empty-blocks
+    constructor(
+        address gatewayAddress,
+        string memory governanceChain_,
+        string memory governanceAddress_,
+        uint256 minimumTimeDelay
+    ) AxelarExecutable(gatewayAddress) TimeLock(minimumTimeDelay) {
+        governanceChainHash = keccak256(bytes(governanceChain_));
+        governanceAddressHash = keccak256(bytes(governanceAddress_));
+    }
+
+    function executeProposal(address targetContract, bytes calldata callData) external virtual {
+        _executeProposal(targetContract, callData);
+    }
+
+    function _executeProposal(address targetContract, bytes calldata callData) internal {
+        bytes32 proposalHash = keccak256(abi.encodePacked(targetContract, callData));
+
+        _executeTimeLock(proposalHash);
+
+        (bool success, ) = targetContract.call(callData);
+
+        if (!success) {
+            revert ExecutionFailed();
+        }
+
+        emit ProposalExecuted(proposalHash);
+    }
+
+    function _execute(
+        string calldata sourceChain,
+        string calldata sourceAddress,
+        bytes calldata payload
+    ) internal override {
+        if (keccak256(bytes(sourceChain)) != governanceChainHash || keccak256(bytes(sourceAddress)) != governanceAddressHash)
+            revert NotGovernance();
+
+        (Command command, address targetContract, bytes memory callData, uint256 eta) = abi.decode(
+            payload,
+            (Command, address, bytes, uint256)
+        );
+
+        if (targetContract == address(0)) revert InvalidTargetContract();
+        if (callData.length == 0) revert InvalidCallData();
+
+        bytes32 proposalHash = keccak256(abi.encodePacked(targetContract, callData));
+
+        if (command == Command.ScheduleProposal) {
+            _scheduleTimeLock(proposalHash, eta);
+
+            emit ProposalScheduled(proposalHash, targetContract, callData, eta);
+        } else if (command == Command.CancelProposal) {
+            _cancelTimeLock(proposalHash);
+
+            emit ProposalCancelled(proposalHash);
+        } else {
+            revert InvalidCommand();
+        }
+    }
+
+    function _executeWithToken(
+        string calldata, /* sourceChain */
+        string calldata, /* sourceAddress */
+        bytes calldata, /* payload */
+        string calldata, /* tokenSymbol */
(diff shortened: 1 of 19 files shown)


contracts/governance/AxelarGovernance.sol
```diff
@@ -0,0 +1,84 @@
...
+    function executeProposal(address target, bytes calldata callData) external payable virtual {
+        _executeProposal(target, callData);
+    }
+
+    function _executeProposal(address target, bytes calldata callData) internal {
+        bytes32 proposalHash = keccak256(abi.encodePacked(target, callData));
+
+        _executeTimeLock(proposalHash);
+
+        (bool success, ) = target.call{ value: msg.value }(callData);
```
milapsheth, line 38: hmm, msg.value should be fixed in the proposal hash when it's scheduled. Otherwise, anyone can frontrun with a different value, e.g. if we are trying to withdraw some tokens etc.

contracts/governance/AxelarGovernance.sol
```diff
@@ -0,0 +1,84 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity 0.8.9;
```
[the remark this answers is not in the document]
re1ro: Should it go into the SDK repo then?

contracts/interfaces/IAxelarGovernance.sol
```diff
@@ -0,0 +1,20 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity 0.8.9;
```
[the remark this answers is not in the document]
re1ro: Should it go into the SDK repo?

contracts/governance/InterchainGovernance.sol
```diff
@@ -0,0 +1,94 @@
...
+        if (command == GovernanceCommand.ScheduleTimeLockProposal) {
+            eta = _scheduleTimeLock(proposalHash, eta);
+
+            emit ProposalScheduled(proposalHash, target, callData, eta);
+        } else if (command == GovernanceCommand.CancelTimeLockProposal) {
+            _cancelTimeLock(proposalHash);
+
+            emit ProposalCancelled(proposalHash);
+        } else {
+            revert InvalidCommand();
```
deanamiel, line 81: I'm not sure the InvalidCommand custom error can ever be hit. It seems like Solidity will throw this panic error instead: "reverted with panic code 0x21 (Tried to convert a value into an enum, but the value was too big or negative)"

contracts/governance/AxelarGovernance.sol
```diff
@@ -0,0 +1,84 @@
...
+
+import { AxelarExecutable } from '@axelar-network/axelar-gmp-sdk-solidity/contracts/executable/AxelarExecutable.sol';
+import { TimeLock } from '@axelar-network/axelar-gmp-sdk-solidity/contracts/utils/TimeLock.sol';
+import { IAxelarGovernance } from '../interfaces/IAxelarGovernance.sol';
+
+contract AxelarGovernance is AxelarExecutable, TimeLock, IAxelarGovernance {
```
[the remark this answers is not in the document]
deanamiel: Added unit tests [here](https://github.com/axelarnetwork/axelar-cgp-solidity/pull/204)

contracts/governance/InterchainGovernance.sol
```diff
@@ -0,0 +1,94 @@
...
+            revert ExecutionFailed();
+        }
+
+        emit ProposalExecuted(proposalHash);
+    }
+
+    function _execute(
+        string calldata sourceChain,
+        string calldata sourceAddress,
+        bytes calldata payload
```
deanamiel, line 46: Probably not a super big deal, but contracts that inherit this module will not be able to encode the payload on-chain as solidity doesn't support implicit conversion from bytes memory to bytes calldata.

contracts/governance/InterchainGovernance.sol
re1ro: Yeah that makes sense. Because we don't have it listed in the enum and it fails in the previous parsing state.

contracts/governance/InterchainGovernance.sol
re1ro: The `_execute` method is always called by a crosschain GMP call and the payload comes from the external call.

contracts/test/TestInterchainGovernance.sol
```diff
@@ -0,0 +1,22 @@
...
+
+contract TestInterchainGovernance is InterchainGovernance {
+    constructor(
+        address gatewayAddress,
+        string memory governanceChain_,
+        string memory governanceAddress_,
+        uint256 minimumTimeDelay
+    ) InterchainGovernance(gatewayAddress, governanceChain_, governanceAddress_, minimumTimeDelay) {}
+
+    function executeProposalAction(
```
milapsheth, line 15: just override the `execute` method instead to skip gateway validation?

contracts/governance/InterchainGovernance.sol
```diff
@@ -0,0 +1,94 @@
...
+        _executeTimeLock(proposalHash);
+
+        (bool success, ) = target.call{ value: msg.value }(callData);
+
+        if (!success) {
+            revert ExecutionFailed();
+        }
+
+        emit ProposalExecuted(proposalHash);
+    }
```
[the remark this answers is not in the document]
re1ro: Good spot!

contracts/test/TestInterchainGovernance.sol
re1ro: execute is not overridable...

contracts/governance/InterchainGovernance.sol
```diff
@@ -18,48 +17,67 @@ contract InterchainGovernance is AxelarExecutable, TimeLock, IInterchainGovernan
...
         governanceAddressHash = keccak256(bytes(governanceAddress_));
     }
 
     /**
-     * @dev Executes a proposal
+     * @notice Returns the ETA of a proposal
+     * @param proposalHash The hash of the proposal
+     * @return uint256 The ETA of the proposal
+     */
+    function getProposalEta(bytes32 proposalHash) external view returns (uint256) {
```
milapsheth, line 49: what i meant was that this getter takes in the same params as executeProposal, and provides an eta for it, vs passing the hash

contracts/governance/InterchainGovernance.sol
milapsheth: do we need the revert then?

contracts/interfaces/IInterchainGovernance.sol
```diff
@@ -4,17 +4,61 @@ pragma solidity ^0.8.0;
...
+
+    /**
+     * @notice Returns the hash of the governance chain.
+     * @return bytes32 The hash of the governance chain
+     */
+    function governanceChainHash() external view returns (bytes32);
+
+    /**
+     * @notice Returns the hash of the governance contract.
+     * @return bytes32 The hash of the governance contract
```
milapsheth, line 42:
```suggestion
     * @notice Returns the hash of the governance address.
     * @return bytes32 The hash of the governance address
```

milapsheth (approved):

deanamiel (approved):

PR #197, opened 2023-06-20
feat(Governance): gateway integration

re1ro:
* [x] `Governance`: Gateway integration
* [x] Depends on #196


contracts/AxelarGateway.sol
```diff
@@ -50,13 +50,21 @@ contract AxelarGateway is IAxelarGateway, AdminMultisigBase {
...
     address internal immutable TOKEN_DEPLOYER_IMPLEMENTATION;
 
-    constructor(address authModule_, address tokenDeployerImplementation_) {
+    constructor(
+        address authModule_,
+        address governanceModule_,
+        address tokenDeployerImplementation_
+    ) {
         if (authModule_.code.length == 0) revert InvalidAuthModule();
+        if (governanceModule_.code.length == 0) revert InvalidGovernanceModule();
```
milapsheth, line 63: Don't think we need to enforce that it's a contract. Could also call it `governance`/`governor` so it feels more generic instead of a module. Also simplifies testing a bit.

contracts/AxelarGateway.sol
```diff
@@ -264,15 +263,7 @@ contract AxelarGateway is IAxelarGateway, AdminMultisigBase {
...
-        (address[] memory adminAddresses, uint256 newAdminThreshold, bytes memory newOperatorsData) = abi.decode(
-            params,
-            (address[], uint256, bytes)
-        );
-
-        // NOTE: Admin epoch is incremented to easily invalidate current admin-related state.
-        uint256 newAdminEpoch = _adminEpoch() + uint256(1);
-        _setAdminEpoch(newAdminEpoch);
-        _setAdmins(newAdminEpoch, adminAddresses, newAdminThreshold);
+        bytes memory newOperatorsData = abi.decode(params, (bytes));
```
milapsheth, line 266: also should be able to set a new governor in setup

contracts/AxelarGateway.sol
sammy1991106: Agree. In that sense, just `GOVERNANCE` instead of `GOVERNANCE_MODULE`.

contracts/AxelarGateway.sol
sammy1991106: Agree, but can do it in a separate PR.

contracts/AxelarGateway.sol
```diff
@@ -11,9 +11,9 @@ import { ITokenDeployer } from './interfaces/ITokenDeployer.sol';
 
 import { ECDSA } from './ECDSA.sol';
 import { DepositHandler } from './DepositHandler.sol';
-import { AdminMultisigBase } from './AdminMultisigBase.sol';
+import { EternalStorage } from './EternalStorage.sol';
 
-contract AxelarGateway is IAxelarGateway, AdminMultisigBase {
+contract AxelarGateway is IAxelarGateway, EternalStorage {
```
sammy1991106, line 16: Unit test

contracts/AxelarGateway.sol
milapsheth: nvm, it's an immutable, so need to deploy a new implementation and upgrade to it to update governance/mint limiter

milapsheth (approved):

contracts/AxelarGateway.sol
```diff
@@ -50,13 +50,21 @@ contract AxelarGateway is IAxelarGateway, AdminMultisigBase {
...
     address internal immutable AUTH_MODULE;
     // solhint-disable-next-line var-name-mixedcase
+    address internal immutable GOVERNANCE;
+    // solhint-disable-next-line var-name-mixedcase
     address internal immutable TOKEN_DEPLOYER_IMPLEMENTATION;
 
-    constructor(address authModule_, address tokenDeployerImplementation_) {
+    constructor(
+        address authModule_,
+        address governance_,
```
deanamiel, line 59: Circular dependency issue with InterchainGovernance constructor may not be ideal

Foivos (approved):

contracts/AxelarGateway.sol
re1ro: Resolved by moving GOVERNANCE into a setup param

milapsheth (approved):

PR #198, opened 2023-06-20
feat(MintLimiter): new module based on Multisig

re1ro:
* [x] `MultisigBase`: new base contract with 
* [x] `MintLimiter`: new module based on Multisig
* [x] Depends on #197


contracts/auth/MultisigBase.sol
```diff
@@ -0,0 +1,111 @@
...
+
+contract MultisigBase is IMultisigBase {
+    struct Voting {
+        uint256 voteCount;
+        mapping(address => bool) hasVoted;
+    }
+
+    struct Signers {
+        address[] accounts;
+        uint256 threshold;
```
milapsheth, line 15: similar comment about not needing to store signer accounts across epochs, only the vote data

contracts/governance/AxelarMultisigMintLimiter.sol
```diff
@@ -0,0 +1,19 @@
...
+import { IMultisigExecute } from '../interfaces/IMultisigExecute.sol';
+import { MultisigBase } from '../auth/MultisigBase.sol';
+
+contract AxelarMultisigMintLimiter is MultisigBase, IMultisigExecute {
+    constructor(address[] memory accounts, uint256 threshold) {
+        _rotateSigners(accounts, threshold);
+    }
+
+    function execute(address target, bytes calldata callData) external payable onlySigners {
+        (bool success, ) = target.call{ value: msg.value }(callData);
```
milapsheth, line 14: msg.value can be influenced by the last signer vs having it as an arg. We can assume the contract was pre-funded with that value to let it go through

contracts/auth/MultisigBase.sol
re1ro: As I've explained the reason in the previous GOV PR. You can't just override a struct in solidity storage. Or we can override the struct manually when rotating operators, but not sure if will save us any gas, because you need to loop and clear existing storage.

contracts/auth/MultisigBase.sol
```diff
@@ -0,0 +1,118 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity 0.8.9;
```
milapsheth, line 3:
```suggestion
pragma solidity ^0.8.0;
```

contracts/governance/AxelarMultisigMintLimiter.sol
```diff
@@ -0,0 +1,32 @@
...
+    constructor(address[] memory accounts, uint256 threshold) {
+        _rotateSigners(accounts, threshold);
+    }
+
+    function execute(
+        address target,
+        bytes calldata callData,
+        uint256 nativeValue
+    ) external payable onlySigners {
+        if (msg.value < nativeValue) revert InsufficientValue();
```
milapsheth, line 21: Similar to the governance contract, I think we should just assume the value is present on the contract's balance already, rather than one specific multisig participant forwarding it. It removes some of the complexity below

contracts/governance/AxelarMultisigMintLimiter.sol
```diff
@@ -0,0 +1,32 @@
...
+
+        (bool success, ) = target.call{ value: nativeValue }(callData);
+        if (!success) {
+            revert ExecutionFailed();
+        }
+
+        if (msg.value > nativeValue) {
```
milapsheth, line 28: not needed with the above change

contracts/interfaces/IMultisigBase.sol
```diff
@@ -0,0 +1,35 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.9;
```
milapsheth, line 3:
```suggestion
pragma solidity ^0.8.0;
```

contracts/governance/AxelarMultisigMintLimiter.sol
```diff
@@ -0,0 +1,32 @@
...
+import { MultisigBase } from '../auth/MultisigBase.sol';
+
+contract AxelarMultisigMintLimiter is MultisigBase, IMultisigExecute {
+    using SafeNativeTransfer for address;
+
+    constructor(address[] memory accounts, uint256 threshold) {
+        _rotateSigners(accounts, threshold);
+    }
+
+    function execute(
```
milapsheth, line 16: Worth adding a separate `MultisigExecute` contract that contains this for reusability (we'll have a need to deploy a standard multisig contract)

contracts/interfaces/IMultisigExecute.sol
```diff
@@ -0,0 +1,16 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.9;
```
milapsheth, line 3:
```suggestion
pragma solidity ^0.8.0;
```

contracts/interfaces/IMultisigBase.sol
```diff
@@ -0,0 +1,35 @@
...
+
+interface IMultisigBase {
```
milapsheth, line 5: Multisig contract should likely be moved to gmp sdk too, but i'm ok with getting it merged first for governance feature completeness

contracts/auth/MultisigBase.sol
milapsheth: ok, we can leave it as-is then

contracts/auth/MultisigBase.sol
```diff
@@ -0,0 +1,118 @@
...
+        // Determine the new vote count.
+        uint256 voteCount = voting.voteCount + 1;
+
+        // Do not proceed with operation execution if insufficient votes.
+        if (voteCount < signers.threshold) {
+            // Save updated vote count.
+            voting.voteCount = voteCount;
+
+            if (msg.value > 0) {
+                msg.sender.safeNativeTransfer(msg.value);
```
milapsheth, line 51: unnecessary complexity, see discussion for `execute`

contracts/governance/AxelarMultisigMintLimiter.sol
```diff
@@ -0,0 +1,32 @@
...
+pragma solidity 0.8.9;
+
+import { SafeNativeTransfer } from '@axelar-network/axelar-gmp-sdk-solidity/contracts/utils/SafeTransfer.sol';
+import { IMultisigExecute } from '../interfaces/IMultisigExecute.sol';
+import { MultisigBase } from '../auth/MultisigBase.sol';
+
+contract AxelarMultisigMintLimiter is MultisigBase, IMultisigExecute {
+    using SafeNativeTransfer for address;
+
+    constructor(address[] memory accounts, uint256 threshold) {
```
milapsheth, line 12: Maybe the constructor should be present in MultisigBase instead to avoid duplication?

deanamiel (approved):

contracts/governance/AxelarMultisigMintLimiter.sol
milapsheth: I'm thinking we can just rename this contract to `Multisig` and it can be used generically for mint limits, etc.

milapsheth (approved):

contracts/AxelarGateway.sol
```diff
@@ -658,4 +670,16 @@ contract AxelarGateway is IAxelarGateway, IGovernable, EternalStorage {
...
         _setAddress(KEY_IMPLEMENTATION, newImplementation);
     }
+
+    function _transferGovernance(address newGovernance) internal {
+        emit GovernanceTransferred(governance(), newGovernance);
+
+        _setAddress(KEY_GOVERNANCE, newGovernance);
+    }
+
+    function _transferMintLimiting(address newMintLimiter) internal {
```
milapsheth, line 680:
```suggestion
    function _transferMintLimiter(address newMintLimiter) internal {
```
feels more consistent naming with transferGovernance?

milapsheth (approved):

PR #199, opened 2023-06-20
feat(MintLimiter): gateway integration

re1ro:
* [x] `MintLimiter`: Gateway integration
* [ ] Depends on #198


contracts/AxelarGateway.sol
```diff
@@ -52,19 +52,24 @@ contract AxelarGateway is IAxelarGateway, EternalStorage {
...
 
     constructor(
         address authModule_,
         address governanceModule_,
+        address mintLimitModule_,
         address tokenDeployerImplementation_
     ) {
         if (authModule_.code.length == 0) revert InvalidAuthModule();
         if (governanceModule_.code.length == 0) revert InvalidGovernanceModule();
+        if (mintLimitModule_.code.length == 0) revert InvalidMintLimiterModule();
```
milapsheth, line 67: similar comment about not requiring this to be a contract

contracts/AxelarGateway.sol
```diff
@@ -178,6 +189,10 @@ contract AxelarGateway is IAxelarGateway, EternalStorage {
         return GOVERNANCE_MODULE;
     }
 
+    function mintLimiterModule() public view returns (address) {
+        return MINT_LIMIT_MODULE;
```
milapsheth, line 193: should allow updating this in setup

milapsheth: nvm, will be done by upgrading implementation since it's an immutable

Foivos (commented):
Ι agree that the IAxelarGateway should not be changed for now, but this is happening in a future PR.

PR #200, opened 2023-06-21
feat: improve gateway deployment script

milapsheth:
- More checks during gateway deployment
- Store gateway addresses in info files
- Allow making axelard queries via an RPC


milapsheth:
use axelarnetwork/axelar-contracts-deployments instead

PR #206, opened 2023-06-27
feat: update solhint to ignore common warnings

deanamiel:
Quick PR to add some common warnings we ignore to .solhint.json.


milapsheth:
Other than AxelarGateway, you can avoid modifying any contracts that have bytecode checks

milapsheth (approved):

PR #208, opened 2023-06-28
use proper filecoin calibration chain id

tzapu:


milapsheth:
thanks, we haven't deployed to filecoin calibration testnet yet. The existing config was for hyperspace testnet which has been deprecated. Once deployed, we'll update the config

milapsheth:
fixed in #230

PR #214, opened 2023-07-11
feat(MultisigBase): signers and voting getters

re1ro:
* [x] `MultisigBase`: signers and voting getters


milapsheth (approved):

deanamiel (approved):
Added tests, MultisigBase.sol is now at full coverage

PR #215, opened 2023-07-14
feat: upgrade gateway test

deanamiel:
PR for governance proposal to upgrade gateway flow


test/GatewayUpgradeFlow.js
milapsheth, line 1: Maybe call it `AxelarGatewayUpgrade.js`

test/GatewayUpgradeFlow.js
```diff
@@ -0,0 +1,180 @@
...
+const {
+    utils: { defaultAbiCoder, Interface, solidityKeccak256 },
+} = ethers;
+const { expect } = chai;
+const { isHardhat, getAddresses, getWeightedAuthDeployParam, getWeightedProxyDeployParams } = require('./utils');
+const { getBytecodeHash } = require('@axelar-network/axelar-contract-deployments');
+
+const getWeights = ({ length }, weight = 1) => Array(length).fill(weight);
+
+describe('InterchainGovernance', () => {
```
milapsheth, line 15:
```suggestion
describe('AxelarGatewayUpgrade', () => {
```

test/GatewayUpgradeFlow.js
```diff
@@ -0,0 +1,180 @@
...
+
+        gatewayFactory = await ethers.getContractFactory('AxelarGateway', owner);
+        authFactory = await ethers.getContractFactory('AxelarAuthWeighted', owner);
+        gatewayProxyFactory = await ethers.getContractFactory('AxelarGatewayProxy', owner);
+        tokenDeployerFactory = await ethers.getContractFactory('TokenDeployer', owner);
+        tokenDeployer = await tokenDeployerFactory.deploy();
+        await tokenDeployer.deployTransaction.wait(network.config.confirmations);
+    });
+
+    beforeEach(async () => {
```
milapsheth, line 56: beforeEach will cause a lot of txs when running on a live network.
Put this in a helper function and group the set of tests in more describe blocks. For e.g. all the queries below can be grouped together, and gateway can be deployed just once to test that (using a before block). See how AxelarGateway test does it

test/GatewayUpgradeFlow.js
```diff
@@ -0,0 +1,180 @@
...
+        const buffer = 10 * 60 * 60;
```
milapsheth, line 57: set it to 6s?

test/GatewayUpgradeFlow.js
```diff
@@ -0,0 +1,180 @@
...
+            .then((d) => d.deployed());
+
+        await gateway.transferGovernance(interchainGovernance.address).then((tx) => tx.wait(network.config.confirmations));
+    });
+
+    it('should get the correct governance address', async () => {
+        expect(await gateway.governance()).to.eq(interchainGovernance.address);
+    });
+
+    it('should get the correct mint limiter address', async () => {
```
milapsheth, line 87: group these tests

test/GatewayUpgradeFlow.js
```diff
@@ -0,0 +1,180 @@
...
+        const nativeValue = 0;
+        const timeDelay = 12 * 60 * 60;
+
+        const targetInterface = new Interface([
+            'function upgrade(address newImplementation, bytes32 newImplementationCodeHash, bytes calldata setupParams) external',
+        ]);
+        const newGatewayImplementation = await gatewayFactory.deploy(auth.address, tokenDeployer.address).then((d) => d.deployed());
+        const newGatewayImplementationCodeHash = await getBytecodeHash(newGatewayImplementation);
+        const setupParams = '0x';
+        const calldata = targetInterface.encodeFunctionData('upgrade', [
```
milapsheth, line 148: why not do target.encodeFunctionDate ...? instead of specifying the ABI explicitly

test/GatewayUpgradeFlow.js
```diff
@@ -0,0 +1,180 @@
...
+    const governanceChain = 'Governance Chain';
+
+    before(async () => {
+        wallets = await ethers.getSigners();
+        ownerWallet = wallets[0];
+        mintLimiter = ownerWallet;
+        governanceAddress = wallets[1].address;
+        operators = sortBy(wallets.slice(0, threshold), (wallet) => wallet.address.toLowerCase());
+
+        interchainGovernanceFactory = await ethers.getContractFactory('TestInterchainGovernance', ownerWallet);
```
milapsheth, line 46: use InterchainGovernance directly, no need for a wrapper

test/GatewayUpgradeFlow.js
```diff
@@ -0,0 +1,180 @@
...
+        const payload = defaultAbiCoder.encode(
+            ['uint256', 'address', 'bytes', 'uint256', 'uint256'],
+            [commandID, target, calldata, nativeValue, eta],
+        );
+
+        await expect(interchainGovernance.executeProposalAction(governanceChain, governanceAddress, payload))
+            .to.emit(interchainGovernance, 'ProposalScheduled')
+            .withArgs(proposalHash, target, calldata, nativeValue, eta);
+
+        await network.provider.send('evm_increaseTime', [timeDelay]);
```
milapsheth, line 168: do a sleep for non hardhat networks

test/GatewayUpgradeFlow.js
deanamiel: Moved gateway deployment into helper function and reorganized tests

test/GatewayUpgradeFlow.js
deanamiel: made it similar to InterchainGovernance tests where buffer is dependent on whether the tests are running on hardhat or a live testnet

test/GatewayUpgradeFlow.js
deanamiel: fixed

test/GatewayUpgradeFlow.js
deanamiel: I changed to define the interface using contract.interface.fragments

test/GatewayUpgradeFlow.js
deanamiel: fixed

test/GatewayUpgradeFlow.js
deanamiel: used waitFor() function similar to InterchainGovernance tests

milapsheth (approved):

test/AxelarGatewayUpgrade.js
```diff
@@ -0,0 +1,180 @@
...
+            .withArgs(proposalHash, target, calldata, nativeValue, eta);
+
+        await waitFor(timeDelay);
+
+        const tx = await interchainGovernance.executeProposal(target, calldata, nativeValue);
+        const executionTimestamp = (await ethers.provider.getBlock(tx.blockNumber)).timestamp;
+
+        await expect(tx)
+            .to.emit(interchainGovernance, 'ProposalExecuted')
+            .withArgs(proposalHash, target, calldata, nativeValue, executionTimestamp)
```
deanamiel, line 176: This test occasionally fails on live testnets where the calculated executionTimestamp differs by a few seconds from the timestamp emitted by the event. Not sure how to get around this.

milapsheth: can you use null or omit the timestamp arg from withArgs?

test/AxelarGatewayUpgrade.js
```diff
@@ -0,0 +1,180 @@
...
+
+        const gatewayImplementation = await gatewayFactory.deploy(auth.address, tokenDeployer.address);
+        await gatewayImplementation.deployTransaction.wait(network.config.confirmations);
+
+        const params = getWeightedProxyDeployParams(ownerWallet.address, mintLimiter.address, [], [], threshold);
+
+        const proxy = await gatewayProxyFactory.deploy(gatewayImplementation.address, params);
+        await proxy.deployTransaction.wait(network.config.confirmations);
+
+        await auth.transferOwnership(proxy.address).then((tx) => tx.wait(network.config.confirmations));
```
milapsheth, line 87:
```suggestion
        awaitConfirmation(auth.transferOwnership(proxy.address));
```
Maybe define a helper `awaitReceipt(tx)` that takes the tx promise and awaits on it and wait for the confirmations. It can import the network in utils, so no need to pass in the confirmation as a param?

test/AxelarGatewayUpgrade.js
deanamiel: I can't emit an arg or put null, a possible work around would be something ugly like checking each event arg from the receipts independently. Something like this:
` const receipt = await tx.wait();
        const event = receipt.events?.find((e) => e.event === 'ProposalExecuted' && e.address === interchainGovernance.address);

        expect(event).to.not.be.undefined;
        expect(event.args.proposalHash).to.equal(proposalHash);
        expect(event.args.target).to.equal(target);
        expect(event.args.callData).to.equal(calldata);
        expect(event.args.value).to.equal(nativeValue);

        if (isHardhat) {
            expect(event.args.timestamp).to.equal(executionTimestamp);
        }` 
        Not sure if we want to to go down this route though.

milapsheth: hmm, i'll play with it a bit. it's weird that timestamp retrieved from block is different than the event

deanamiel: sounds good, let me know

test/AxelarGatewayUpgrade.js
```diff
@@ -0,0 +1,181 @@
...
+        expect(await gateway.authModule()).to.eq(auth.address);
+        expect(await auth.owner()).to.eq(gateway.address);
+        expect(await gateway.tokenDeployer()).to.eq(tokenDeployer.address);
+    });
+
+    it('should upgrade AxelarGateway through InterchainGovernance proposal', async () => {
+        const commandID = 0;
+        const target = gateway.address;
+        const nativeValue = 0;
+        const timeDelay = isHardhat ? 12 * 60 * 60 : 20;
```
milapsheth, line 110:
```suggestion
        const timeDelay = isHardhat ? 12 * 60 * 60 : 15;
```

PR #219, opened 2023-08-01
feat: gateway test coverage

deanamiel:
PR to expand test coverage for AxelarGateway.


test/AxelarGateway.js
```diff
@@ -269,21 +349,74 @@ describe('AxelarGateway', () => {
...
+                .withArgs(newGatewayImplementation.address)
+                .to.emit(gateway, 'GovernanceTransferred')
+                .withArgs(governance.address, notGovernance.address)
+                .to.emit(gateway, 'MintLimiterTransferred')
+                .withArgs(governance.address, notGovernance.address);
+        });
+
+        it('should allow governance to upgrade to the correct implementation with new operators', async () => {
+            const newGatewayImplementation = await gatewayFactory.deploy(auth.address, tokenDeployer.address).then((d) => d.deployed());
+            const newGatewayImplementationCodeHash = await getBytecodeHash(newGatewayImplementation);
```
milapsheth, line 394:
```suggestion
            const newGatewayImplementationCodeHash = await getBytecodeHash(newGatewayImplementation, network.config.id);
```
can you add this everywhere this method is used?

test/AxelarGateway.js
```diff
@@ -671,276 +828,409 @@ describe('AxelarGateway', () => {
...
+                await deployGateway();
+            });
 
-            const data = buildCommandBatch(
-                await getChainId(),
-                [getRandomID()],
-                ['transferOperatorship'],
-                [getTransferWeightedOperatorshipCommand(newOperators, getWeights(newOperators), newOperators.length)],
-            );
+            before(async () => {
```
milapsheth, line 1093: maybe create a helper under this test? only difference with positive test is the different factory

test/AxelarGateway.js
```diff
@@ -1156,105 +1450,163 @@ describe('AxelarGateway', () => {
...
+                await gateway.execute(input, getGasOptions()).then((tx) => tx.wait());
+
+                const tokenAddress = await gateway.tokenAddresses(tokenSymbol);
+                expect(tokenAddress).to.equal(AddressZero);
+            });
         });
 
-        it('should burn internal token and emit an event', async () => {
-            const tokenSymbol = `TEST${getRandomString(10)}`;
+        describe('valid token deployer', () => {
```
milapsheth, line 1483:
```suggestion
        describe('send token', () => {
```
this is testing send token it seems

test/AxelarGateway.js
```diff
@@ -1372,6 +1763,91 @@ describe('AxelarGateway', () => {
...
+
+            expect(values).to.deep.eq([name, symbol, decimals, cap, amount1, amount2]);
+        });
+
+        it('should not execute the same commandID twice', async () => {
+            const duplicateCommandID = getRandomID();
+            const data = buildCommandBatch(
+                await getChainId(),
+                [getRandomID(), getRandomID(), duplicateCommandID, duplicateCommandID],
+                ['deployToken', 'mintToken', 'mintToken', 'transferOperatorship'],
```
milapsheth, line 1813: I don't see why the other commands are needed here for testing duplicate checks. Also, let's avoid using expensive commands like token deployments unless the test is focused on that. I recommend using approveContractCall instead since it's the cheapest and doesn't require a token to be setup for the test. This could simplify other tests too.

And feel free to create helper functions where applicable to reduce code duplication

test/AxelarGateway.js
```diff
@@ -1401,6 +1877,50 @@ describe('AxelarGateway', () => {
...
+            );
+        });
+
+        it('should revert if token amount is invalid', async () => {
+            const tokenSymbol = getRandomString(10);
+            const data = buildCommandBatch(
+                await getChainId(),
+                [getRandomID()],
+                ['deployToken'],
+                [getDeployCommand(tokenName, tokenSymbol, decimals, cap, ethers.constants.AddressZero, 0)],
```
milapsheth, line 1900: instead of deploying a new token, you could register an existing test token (that can be deployed exactly once before all tests). deployToken for an existing token is much cheaper

test/AxelarGateway.js
```diff
@@ -1650,4 +2214,33 @@ describe('AxelarGateway', () => {
             expect(isApprovedAfter).to.be.false;
         });
     });
+
+    describe('deprecated functions', () => {
```
milapsheth, line 2218: sweet

test/AxelarGateway.js
```diff
@@ -1650,4 +2214,33 @@ describe('AxelarGateway', () => {
...
+        it('should return correct value for allTokensFrozen', async () => {
+            expect(await gateway.allTokensFrozen()).to.be.false;
+        });
+
+        it('should return correct value for adminEpoch', async () => {
+            expect(await gateway.adminEpoch()).to.eq(0);
+        });
+
+        it('should return correct value for adminThreshold', async () => {
+            const epoch = 1;
```
milapsheth, line 2232: try using random ints for better coverage

contracts/test/TestMintableCappedERC20.sol
```diff
@@ -0,0 +1,31 @@
...
+    ) MintableCappedERC20(name, symbol, decimals, capacity) {
+        shouldFailTransfer = false;
+    }
+
+    function setFailTransfer(bool _shouldFailTransfer) public {
+        shouldFailTransfer = _shouldFailTransfer;
+    }
+
+    function transfer(address recipient, uint256 amount) public view override returns (bool) {
+        if (shouldFailTransfer) {
```
milapsheth, line 26: is this extra contract needed? you can trigger a failure in the original contract using 0 address for recipient

test/AxelarGateway.js
```diff
@@ -370,6 +503,18 @@ describe('AxelarGateway', () => {
             await expect(implementation.connect(governance).setup(params)).to.be.revertedWithCustomError(implementation, 'NotProxy');
         });
 
+        it('should not allow malicious proxy to call setup function directly and transfer governance or mint limiter', async () => {
```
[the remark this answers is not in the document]
deanamiel: Simulates someone deploying their own instance of the gateway proxy, pointing it to the existing gateway implementation, and passing in setup params to transfer governance/mint limiter to a different address

contracts/test/TestMintableCappedERC20.sol
deanamiel: I couldn't think of a way around this contract. The recipient is hardcoded as `address(this)` referring to the gateway itself. Even if I could make the recipient the 0 address the transfer would revert and I would only be able to cover the first part of the `or` condition below:
` if (!success || (returnData.length != uint256(0) && !abi.decode(returnData, (bool)))) revert BurnFailed(symbol);`
Line 446

test/AxelarGateway.js
deanamiel: I replaced all the commands with approveContractCall to save gas (since all other commands are tested separately anyways), I also combined two of the tests into one because I realized there was a coverage overlap. These changes allowed me to turn the beforeEach to a before for deploying the gateway. Ran the batch commands on fantom and they were significantly faster.

contracts/test/TestMintableCappedERC20.sol
re1ro: Some ERC20 token implementations just return `false` when transfer fails instead of reverting. Maybe you would need to add such an ERC20 mock to cover that

deanamiel: Yeah, I like that name better, just changed it

test/AxelarGateway.js
milapsheth: you can call it "negative send token test" maybe? Or since it's only one, we don't really need a separate describe block. Just feel like invalid token deployer isn't accurate in terms of what's being tested.

test/AxelarGateway.js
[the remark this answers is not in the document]
milapsheth: yeah, but i don't see setup etc. being called in the test?

test/AxelarGateway.js
```diff
@@ -2215,12 +2152,12 @@ describe('AxelarGateway', () => {
         });
 
         it('should return correct value for adminThreshold', async () => {
-            const epoch = 1;
+            const epoch = getRandomInt(Number.MAX_SAFE_INTEGER);
             expect(await gateway.adminThreshold(epoch)).to.eq(0);
```
milapsheth, line 2156: thought about this again, and maybe it's worth also explicitly verifying for the epoch = 1 case in both tests, since we always start with that.

test/AxelarGateway.js
```diff
@@ -970,10 +971,10 @@ describe('AxelarGateway', () => {
                 const salt = id(`${destinationAddress}-${owner.address}-${getRandomInt(1e10)}`);
                 const depositHandlerAddress = getCreate2Address(gateway.address, salt, keccak256(depositHandlerFactory.bytecode));
 
-                await wallets[0].sendTransaction({ to: depositHandlerAddress, value: '100000000000000000' });
+                await wallets[0].sendTransaction({ to: depositHandlerAddress, value: '100000000000000000' }).then((tx) => tx.wait());
```
milapsheth, line 974:
```suggestion
                await wallets[0].sendTransaction({ to: depositHandlerAddress, value: '1' }).then((tx) => tx.wait());
```
huge value here. can you sanity check other occurrences in all our tests?
instead of wallets[0], refer to a specific wallet?

test/AxelarGateway.js
milapsheth: any reason to not refactor this into a helper?

test/AxelarGateway.js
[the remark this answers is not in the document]
deanamiel: Since it's just another copy of AxelarGatewayProxy setup is called in the constructor when it's deployed. I could make a test proxy that calls setup more explicitly or remove the test as it doesn't add to test coverage at all. I just saw some people bring up this question in code4rena so I thought I might as well add a test for it.

test/AxelarGateway.js
```diff
@@ -1658,28 +1668,25 @@ describe('AxelarGateway', () => {
...
 
         it('should revert on mismatch between commandID and command/params length', async () => {
+            const payload = defaultAbiCoder.encode(['address'], [owner.address]);
+            const payloadHash = keccak256(payload);
             let data = buildCommandBatch(
                 await getChainId(),
-                [getRandomID(), getRandomID(), getRandomID(), getRandomID()],
-                ['deployToken', 'mintToken', 'mintToken'],
+                [getRandomID(), getRandomID(), getRandomID()],
+                ['approveContractCall', 'approveContractCall'],
```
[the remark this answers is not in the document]
deanamiel: yeah more concise

test/AxelarGateway.js
deanamiel: added epoch=1 and then a random number test for both tests

test/AxelarGateway.js
deanamiel: Replaced all instances of wallets[0] with owner for readability. This appears to be the only place where explicit value is passed in with a call, so changed it to '1' as per your suggestion.

test/AxelarGateway.js
deanamiel: I deployed an external token in the before block and used it in a few tests, there were still 2 tests that were having issue using this shared external token, so I kept those tests the same.

test/AxelarGateway.js
```diff
@@ -1953,16 +1890,16 @@ describe('AxelarGateway', () => {
         });
 
         it('should lock external token and emit an event', async () => {
-            const tokenSymbol = getRandomString(10);
-            const token = await mintableCappedERC20Factory.deploy(tokenName, tokenSymbol, decimals, cap).then((d) => d.deployed());
+            const externalTokenSymbol = getRandomString(10);
+            const token = await mintableCappedERC20Factory.deploy(tokenName, externalTokenSymbol, decimals, cap).then((d) => d.deployed());
```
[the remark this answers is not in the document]
deanamiel: I deployed an external token in the before block and used it in a few tests, there were still 2 tests that were having issue using this shared external token, so I kept those tests the same.

test/AxelarGateway.js
```diff
@@ -71,6 +77,15 @@ describe('AxelarGateway', () => {
         // reuse token deployer for all tests
         tokenDeployer = await tokenDeployerFactory.deploy();
         await tokenDeployer.deployTransaction.wait(network.config.confirmations);
+
+        // reuse external token for all tests
+        externalName = 'An External Token';
+        externalSymbol = 'AET';
+        externalDecimals = 18;
+        externalCap = 1e8;
```
milapsheth, line 85:
```suggestion
        externalCap = 0;
```
don't cap it

milapsheth, line 85:
```suggestion
        externalTokenSymbol = 'AET';
        externalTokenDecimals = 18;
        externalTokenCap = 0;
```
don't cap it

test/AxelarGateway.js
```diff
@@ -71,6 +77,15 @@ describe('AxelarGateway', () => {
         // reuse token deployer for all tests
         tokenDeployer = await tokenDeployerFactory.deploy();
         await tokenDeployer.deployTransaction.wait(network.config.confirmations);
+
+        // reuse external token for all tests
+        externalName = 'An External Token';
```
milapsheth, line 82:
```suggestion
        externalTokenName = 'An External Token';
```

test/AxelarGateway.js
deanamiel: I can add a quick fix to refactor the common code into a helper function

test/AxelarGateway.js
```diff
@@ -1890,16 +1895,17 @@ describe('AxelarGateway', () => {
         });
 
         it('should lock external token and emit an event', async () => {
-            const externalTokenSymbol = getRandomString(10);
-            const token = await mintableCappedERC20Factory.deploy(tokenName, externalTokenSymbol, decimals, cap).then((d) => d.deployed());
+            const externalTokenSymbol = externalSymbol;
+            const token = externalToken;
```
milapsheth, line 1899: unnecessary vars?

deanamiel: fixed

test/AxelarGateway.js
deanamiel: fixed

milapsheth (approved):

re1ro (approved):

PR #220, opened 2023-08-02
feat(Governance): dedicated withdraw method and safety checks

re1ro:
* [x] `Governance`: dedicated withdraw method that can only be called via proposal
* [x] `Caller`: checking if target is a contract
* [x]  `Governance`: `_getProposalHash` to use `abi.encode`


contracts/governance/InterchainGovernance.sol
```diff
@@ -70,14 +94,24 @@ contract InterchainGovernance is AxelarExecutable, TimeLock, Caller, IInterchain
...
     }
 
+    /**
+     * @notice Withdraws native token from the contract
+     * @param to The address to send the native token to
+     * @param amount The amount of native token to send
+     * @dev This function is only callable by the contract itself after passing according proposal
+     */
+    function withdraw(address to, uint256 amount) external onlySelf {
+        to.safeNativeTransfer(amount);
```
milapsheth, line 112:
```suggestion
     * @param recipient The address to send the native token to
     * @param amount The amount of native token to send
     * @dev This function is only callable by the contract itself after passing according proposal
     */
    function withdraw(address recipient, uint256 amount) external onlySelf {
        recipient.safeNativeTransfer(amount);
```

contracts/interfaces/IInterchainGovernance.sol
```diff
@@ -67,4 +68,12 @@ interface IInterchainGovernance is IAxelarExecutable, ICaller {
...
     ) external payable;
+
+    /**
+     * @notice Withdraws native token from the contract
+     * @param to The address to send the native token to
+     * @param amount The amount of native token to send
+     * @dev This function is only callable by the contract itself after passing according proposal
+     */
+    function withdraw(address to, uint256 amount) external;
 }
```
milapsheth, line 79:
```suggestion
     * @notice Withdraws native token from the contract
     * @param recipient The address to send the native token to
     * @param amount The amount of native token to send
     * @dev This function is only callable by the contract itself after passing according proposal
     */
    function withdraw(address recipient, uint256 amount) external;
}
```

milapsheth (approved):

contracts/governance/AxelarServiceGovernance.sol
```diff
@@ -81,7 +81,7 @@ contract AxelarServiceGovernance is InterchainGovernance, MultisigBase, IAxelarS
         }
```
deanamiel, line 81: With the check above the last else-if can just become an else because we know at that point the commandID must be 3 or it would have reverted. Right now with the else-if test coverage cannot reach 100% because it wants to check the case where the last condition fails, but that is not possible. Unless we want to keep it like this for readability.

milapsheth (approved):

test/InterchainGovernance.js
```diff
@@ -151,39 +125,67 @@ describe('InterchainGovernance', () => {
...
-        const commandIDCancel = 1;
-        const target = targetContract.address;
+        const target = interchainGovernance.address;
         const nativeValue = 100;
-        const timeDelay = 12 * 60 * 60;
+        const recipient = ownerWallet.address;
 
-        const targetInterface = new Interface(['function callTarget() external']);
-        const calldata = targetInterface.encodeFunctionData('callTarget');
+        const withdrawInterface = new Interface(['function withdraw(address recipient, uint256 amount) external']);
```
milapsheth, line 154: why hardcode the interface again? should create it dynamically
https://github.com/axelarnetwork/axelar-gmp-sdk-solidity/blob/main/test/Operators.js#L154

test/InterchainGovernance.js
```diff
@@ -199,27 +201,17 @@ describe('InterchainGovernance', () => {
...
+        await interchainGovernance.executeProposalAction(governanceChain, governanceAddress.address, payload).then((tx) => tx.wait());
 
-        await network.provider.send('evm_increaseTime', [timeDelay]);
-        await network.provider.send('evm_mine');
+        if (isHardhat) {
+            await network.provider.send('evm_increaseTime', [timeDelay]);
+            await network.provider.send('evm_mine');
+        } else {
+            await new Promise((resolve) => setTimeout(resolve, timeDelay * 1000));
+        }
```
milapsheth, line 214: duplicated code everywhere, use a helper

test/InterchainGovernance.js
```diff
@@ -151,39 +125,67 @@ describe('InterchainGovernance', () => {
...
-            .to.emit(interchainGovernance, 'ProposalScheduled')
-            .withArgs(proposalHash, target, calldata, nativeValue, eta);
+        if (isHardhat) {
+            await network.provider.send('evm_increaseTime', [timeDelay]);
+            await network.provider.send('evm_mine');
+        } else {
+            await new Promise((resolve) => setTimeout(resolve, timeDelay * 1000));
+        }
+
+        const tx = await interchainGovernance.executeProposal(target, withdrawCalldata, nativeValue);
```
milapsheth, line 175:
```suggestion
        const tx = await interchainGovernance.executeProposal(target, withdrawCalldata, 0);
```

test/InterchainGovernance.js
```diff
@@ -151,39 +125,67 @@ describe('InterchainGovernance', () => {
...
+        const recipient = ownerWallet.address;
 
-        const targetInterface = new Interface(['function callTarget() external']);
-        const calldata = targetInterface.encodeFunctionData('callTarget');
+        const withdrawInterface = new Interface(['function withdraw(address recipient, uint256 amount) external']);
+        const withdrawCalldata = withdrawInterface.encodeFunctionData('withdraw', [recipient, nativeValue]);
 
-        const block = await ethers.provider.getBlock('latest');
-        const eta = block.timestamp + timeDelay;
+        const [payload] = await getPayloadAndProposalHash(commandID, target, nativeValue, withdrawCalldata, timeDelay);
```
milapsheth, line 157:
```suggestion
        const [payload] = await getPayloadAndProposalHash(commandID, target, 0, withdrawCalldata, timeDelay);
```
nativeValue is `0`. The target call here is `withdraw`. We aren't forwarding any native ether as msg.value to this call. `withdraw` will withdraw ether for us instead.

contracts/governance/AxelarServiceGovernance.sol
milapsheth: how about instead of casting `commandId` to an enum, we convert the enums to uint256 during the if comparison, and thus add an `else` clause that reverts. We won't need the explicit check for greater max enum value anymore and it's easier for coverage and slither to reason about.

test/InterchainGovernance.js
deanamiel: updated

test/InterchainGovernance.js
deanamiel: updated

contracts/governance/AxelarServiceGovernance.sol
deanamiel: I implemented the change, let me know if this is what you were referring to

test/utils.js
```diff
@@ -57,6 +61,15 @@ const getPayloadAndProposalHash = async (commandID, target, nativeValue, calldat
     return [payload, proposalHash, eta];
 };
 
+const increaseTime = async (timeDelay) => {
```
milapsheth, line 64:
```suggestion
const waitFor = async (timeDelay) => {
```
a bit more readable

test/utils.js
```diff
@@ -18,6 +18,10 @@ const getRandomString = (length) => {
 
 const getAddresses = (wallets) => wallets.map(({ address }) => address);
 
+const isHardhat = () => {
+    return network.name === 'hardhat';
+};
```
milapsheth, line 23:
```suggestion
const isHardhat = network.name === 'hardhat';
```
doesn't need to be a function

test/utils.js
```diff
@@ -57,6 +61,15 @@ const getPayloadAndProposalHash = async (commandID, target, nativeValue, calldat
     return [payload, proposalHash, eta];
 };
 
+const increaseTime = async (timeDelay) => {
+    if (isHardhat) {
```
milapsheth, line 65: isHardhat is currently a function, and it isn't being called here. but doesn't need to be a function

test/utils.js
deanamiel: updated

test/utils.js
deanamiel: yup changed to bool

deanamiel (approved):

PR #225, opened 2023-08-10
feat: improve live revert tests

milapsheth:
Some chains like mantle aren't able to pass custom revert tests based on rpc simulation failures. This PR enables a workaround for them

### DIFF
--- a/contracts/governance/InterchainGovernance.sol
+++ b/contracts/governance/InterchainGovernance.sol
@@ -191,7 +191,7 @@ contract InterchainGovernance is AxelarExecutable, TimeLock, Caller, IInterchain
     }
 
     /**
-     * @notice Making contact able to receive native value
+     * @notice Allow contract to receive native token
      */
     receive() external payable {}
 }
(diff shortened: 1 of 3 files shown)


deanamiel (approved):

contracts/governance/InterchainGovernance.sol
```diff
@@ -191,7 +191,7 @@ contract InterchainGovernance is AxelarExecutable, TimeLock, Caller, IInterchain
     }
 
     /**
-     * @notice Making contact able to receive native value
+     * @notice Allow contract to receive native token
```
re1ro, line 194: Maybe native coin/ether/gas token? Token usually means ECR20, so I am not sure about this.

re1ro (approved):

contracts/governance/InterchainGovernance.sol
milapsheth, line 194:
```suggestion
     * @notice Allow contract to receive native gas token
```

milapsheth: Sure, fixed

PR #227, opened 2023-08-10
fix(MultisigBase): gas usage optimizations

re1ro:
* [x] MultisigBase: gas usage optimizations

### DIFF
--- a/contracts/auth/MultisigBase.sol
+++ b/contracts/auth/MultisigBase.sol
@@ -63,12 +63,12 @@ contract MultisigBase is IMultisigBase {
         }
 
         // Clear vote count and voted booleans.
-        voting.voteCount = 0;
+        delete voting.voteCount;
 
         uint256 count = signers.accounts.length;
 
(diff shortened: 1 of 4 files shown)


deanamiel (approved):

contracts/auth/MultisigBase.sol
```diff
@@ -42,37 +42,7 @@ contract MultisigBase is IMultisigBase {
...
-
-        uint256 count = signers.accounts.length;
-
-        for (uint256 i; i < count; ++i) {
-            voting.hasVoted[signers.accounts[i]] = false;
-        }
-
-        emit MultisigOperationExecuted(topic);
-
+        if (_onlySigners()) return;
```
milapsheth, line 45:
```suggestion
        if (!_onlySigners()) return;
```
switch the return value around so that the naming makes more sense?

deanamiel: Yeah makes sense, will change the name to _isFinalSignerVote too as per @re1ro's suggestion

PR #228, opened 2023-08-10
refactor: remove governance contracts and tests

deanamiel:
PR to remove interchain governance from CGP repo.

Note: AxelarGatewayUpgrade test will fail until new version of SDK is released because it has InterchainGovernance ABI as a dependency.


milapsheth (approved):

milapsheth:
Point to the governance design doc in the gmp sdk instead of keeping the duplicate

milapsheth (approved):

PR #232, opened 2023-08-16
feat: use sdk ownable and upgradable

deanamiel:
PR to use SDK ownable and upgradable contracts in CGP repo.


contracts/auth/AxelarAuthWeighted.sol
```diff
@@ -13,7 +13,7 @@ contract AxelarAuthWeighted is Ownable, IAxelarAuthWeighted {
 
     uint256 internal constant OLD_KEY_RETENTION = 16;
 
-    constructor(bytes[] memory recentOperators) {
+    constructor(bytes[] memory recentOperators, address owner) Ownable(owner) {
```
milapsheth, line 16:
```suggestion
    constructor(bytes[] memory recentOperators) Ownable(msg.sender) {
```
avoids breaking the deployment this way, and it can be transferred later as usual

contracts/interfaces/IAxelarAuth.sol
```diff
@@ -2,9 +2,7 @@
 
 pragma solidity ^0.8.9;
```
milapsheth, line 3:
```suggestion
pragma solidity ^0.8.0;
```
change it for other AxelarAuth interface and contract too since we're redeploying those

contracts/interfaces/IAxelarAuth.sol
```diff
@@ -2,9 +2,7 @@
...
 
-import { IOwnable } from './IOwnable.sol';
-
-interface IAxelarAuth is IOwnable {
+interface IAxelarAuth {
```
milapsheth, line 5: inherit from gmp sdk IOwnable?

contracts/interfaces/IAxelarDepositService.sol
```diff
@@ -2,10 +2,9 @@
 
 pragma solidity ^0.8.0;
 
-import { IUpgradable } from './IUpgradable.sol';
 import { IDepositServiceBase } from './IDepositServiceBase.sol';
 
-interface IAxelarDepositService is IUpgradable, IDepositServiceBase {
+interface IAxelarDepositService is IDepositServiceBase {
```
milapsheth, line 7: inherit from gmp sdk IUpgradable?

contracts/interfaces/IAxelarGasService.sol
```diff
@@ -2,9 +2,7 @@
 
 pragma solidity ^0.8.0;
 
-import { IUpgradable } from './IUpgradable.sol';
-
-interface IAxelarGasService is IUpgradable {
+interface IAxelarGasService {
```
milapsheth, line 5: same

contracts/interfaces/IAxelarAuth.sol
deanamiel: Good point

contracts/interfaces/IAxelarDepositService.sol
deanamiel: fixed

contracts/interfaces/IAxelarGasService.sol
deanamiel: fixed

milapsheth (approved):

PR #256, opened 2023-10-26
fix: gateway upgrade test

milapsheth:


test/AxelarGatewayUpgrade.js
```diff
@@ -179,7 +178,7 @@ describe('AxelarGatewayUpgrade', () => {
             .to.emit(interchainGovernance, 'ProposalScheduled')
             .withArgs(proposalHash, target, calldata, nativeValue, finalEta);
 
-        await waitFor(timeDelay);
+        await waitFor(timeDelay, () => {});
```
milapsheth, line 181: seems to be a regression, callback should be optional @blockchainguyy

PR #259, opened 2023-11-21
feat: check bytecode size in gh actions

milapsheth:


deanamiel (approved):

.github/workflows/test.yaml
```diff
@@ -23,5 +23,8 @@ jobs:
       - name: Install Dependencies
         run: npm ci
 
+      - name: Set environment variable
+        run: export CHECK_CONTRACT_SIZE=true
```
re1ro, line 27: Would it make more sense to set the env var only for the following test command instead of exporting it into the env? Something like
```
      - name: Test
        run: CHECK_CONTRACT_SIZE=true npm run test-evm-versions
```

re1ro (approved):
LGTM, left one suggestion

.github/workflows/test.yaml
milapsheth: true, will change that in other repos as well

PR #270, opened 2024-02-05
refactor: update eth_estimateGas test comment

blockchainguyy:
Jira: https://axelarnetwork.atlassian.net/browse/AXE-3064

Update the comment to explain that this test is to check if the gas requirements for a new chain align with the Ethereum requirements.
Added a test to check transaction goes through using estimated gas.

### DIFF
--- a/test/RpcCompatibility.js
+++ b/test/RpcCompatibility.js
@@ -252,7 +252,7 @@ describe('RpcCompatibility', () => {
 
         expect(estimatedGas).to.be.a('string');
         expect(gas).to.be.gt(0);
-        expect(gas).to.be.lt(30000); // report if gas estimation is different than ethereum
+        expect(gas).to.be.lt(30000); // report if gas estimation matches Ethereum's behavior to adjust core configuration if necessary.
     });
 
     it('should support RPC method eth_gasPrice', async () => {
--- a/test/RpcCompatibility.js
+++ b/test/RpcCompatibility.js
@@ -252,7 +252,7 @@ describe('RpcCompatibility', () => {
 
         expect(estimatedGas).to.be.a('string');
         expect(gas).to.be.gt(0);
-        expect(gas).to.be.lt(30000); // report if gas estimation matches Ethereum's behavior to adjust core configuration if necessary.
+        expect(gas).to.be.lt(30000); // report if gas estimation does not matches Ethereum's behavior to adjust core configuration if necessary.
     });
 
     it('should support RPC method eth_gasPrice', async () => {
--- a/test/RpcCompatibility.js
+++ b/test/RpcCompatibility.js
@@ -255,6 +255,18 @@ describe('RpcCompatibility', () => {
         expect(gas).to.be.lt(30000); // report if gas estimation does not matches Ethereum's behavior to adjust core configuration if necessary.
     });
 
+    it('should send tx with estimated gas', async () => {
+        const newValue = 300;
+        const txParams = {
+            to: rpcCompatibilityContract.address,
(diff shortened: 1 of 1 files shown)


test/RpcCompatibility.js
```diff
@@ -252,7 +252,19 @@ describe('RpcCompatibility', () => {
...
+
+    it('should send tx with estimated gas', async () => {
+        const newValue = 300;
+        const txParams = {
+            to: rpcCompatibilityContract.address,
+            data: rpcCompatibilityContract.interface.encodeFunctionData('updateValue', [newValue]),
+        };
+
+        const estimatedGas = await provider.estimateGas(txParams);
+        const receipt = await rpcCompatibilityContract.updateValue(newValue, { gasLimit: estimatedGas }).then((tx) => tx.wait());
```
milapsheth, line 266: when updateValue is called without gas limit, ethers already estimates the gas needed so this should already be covered. but i'm fine with being explicit in a test.

test/RpcCompatibility.js
```diff
@@ -252,7 +252,19 @@ describe('RpcCompatibility', () => {
...
+    });
+
+    it('should send tx with estimated gas', async () => {
+        const newValue = 300;
+        const txParams = {
+            to: rpcCompatibilityContract.address,
+            data: rpcCompatibilityContract.interface.encodeFunctionData('updateValue', [newValue]),
+        };
+
+        const estimatedGas = await provider.estimateGas(txParams);
```
milapsheth, line 265: there's probably a better to call estimateGas on a populated tx instead of populating it yourself above

milapsheth:
JIRA task identifier is AXE and not AXL. We don't really put jira identifiers in commits atm. Can you add it in the PR description instead for consistency

test/RpcCompatibility.js
```diff
@@ -240,19 +240,31 @@ describe('RpcCompatibility', () => {
...
+            const gas = BigNumber.from(estimatedGas);
+
+            expect(estimatedGas).to.be.a('string');
+            expect(gas).to.be.gt(0);
+            expect(gas).to.be.lt(30000); // report if gas estimation does not matches Ethereum's behavior to adjust core configuration if necessary.
+        });
+
+        it('should send tx with estimated gas', async () => {
+            const newValue = 300;
+            const tx = await signer.populateTransaction(await rpcCompatibilityContract.populateTransaction.updateValue(newValue));
```
milapsheth, line 261:
```suggestion
            const tx = await rpcCompatibilityContract.connect(signer).populateTransaction.updateValue(newValue);
```
does this populate all tx fields instead?

blockchainguyy: Yes, you can see the diff here:

```sh
tx 1 {
  data: '0x573c0bd3000000000000000000000000000000000000000000000000000000000000012c',
  to: '0x5FbDB2315678afecb367f032d93F642f64180aa3',
  from: '0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266',
  gasLimit: BigNumber { value: "29021528" },
  type: 2,
  maxFeePerGas: BigNumber { value: "3250000000" },
  maxPriorityFeePerGas: BigNumber { value: "1500000000" },
  nonce: 1,
  chainId: 31337
}
tx2 {
  data: '0x573c0bd3000000000000000000000000000000000000000000000000000000000000012c',
  to: '0x5FbDB2315678afecb367f032d93F642f64180aa3',
  from: '0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266',
  gasLimit: BigNumber { value: "29021528" }
}
```

re1ro (approved):

PR #279, opened 2024-04-17
feat(GasService): updated estimation dependency

re1ro:
* [x] `AxelarGasService`: updating integration with the `GasEstimation` dependency
* [x] Depends on the GMP-SDK release

### DIFF
--- a/contracts/gas-service/AxelarGasService.sol
+++ b/contracts/gas-service/AxelarGasService.sol
@@ -4,7 +4,7 @@ pragma solidity ^0.8.0;
 
 import { IERC20 } from '@axelar-network/axelar-gmp-sdk-solidity/contracts/interfaces/IERC20.sol';
 import { IAxelarGasService } from '@axelar-network/axelar-gmp-sdk-solidity/contracts/interfaces/IAxelarGasService.sol';
-import { InterchainGasEstimation } from '@axelar-network/axelar-gmp-sdk-solidity/contracts/gas-estimation/InterchainGasEstimation.sol';
+import { InterchainGasEstimation, GasInfo } from '@axelar-network/axelar-gmp-sdk-solidity/contracts/gas-estimation/InterchainGasEstimation.sol';
 import { Upgradable } from '@axelar-network/axelar-gmp-sdk-solidity/contracts/upgradable/Upgradable.sol';
 import { SafeTokenTransfer, SafeTokenTransferFrom } from '@axelar-network/axelar-gmp-sdk-solidity/contracts/libs/SafeTransfer.sol';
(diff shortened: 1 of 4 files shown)


contracts/gas-service/AxelarGasService.sol
```diff
@@ -4,7 +4,7 @@ pragma solidity ^0.8.0;
 
 import { IERC20 } from '@axelar-network/axelar-gmp-sdk-solidity/contracts/interfaces/IERC20.sol';
 import { IAxelarGasService } from '@axelar-network/axelar-gmp-sdk-solidity/contracts/interfaces/IAxelarGasService.sol';
-import { InterchainGasEstimation } from '@axelar-network/axelar-gmp-sdk-solidity/contracts/gas-estimation/InterchainGasEstimation.sol';
+import { InterchainGasEstimation, GasInfo } from '@axelar-network/axelar-gmp-sdk-solidity/contracts/gas-estimation/InterchainGasEstimation.sol';
```
milapsheth, line 7: hmm, why didn't it fail to compile before?

milapsheth (approved):

contracts/gas-service/AxelarGasService.sol
```diff
@@ -4,7 +4,7 @@ pragma solidity ^0.8.0;
```
milapsheth, line 4: bump package version?

contracts/gas-service/AxelarGasService.sol
re1ro: Because `GasInfo` was defined in the interface.

contracts/gas-service/AxelarGasService.sol
re1ro: +1

contracts/gas-service/AxelarGasService.sol
```diff
@@ -4,7 +4,7 @@ pragma solidity ^0.8.0;
 
 import { IERC20 } from '@axelar-network/axelar-gmp-sdk-solidity/contracts/interfaces/IERC20.sol';
```
milapsheth, line 5: need to bump gmp sdk release

re1ro: In package.json? Not released yet, right?

contracts/gas-service/AxelarGasService.sol
milapsheth: ah, makes sense

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@axelar-network/axelar-cgp-solidity",
-  "version": "6.3.0",
+  "version": "6.3.1",
```
milapsheth, line 3: will need to cherry pick this commit to the release branch for the patch release. or fine with a minor version bump since we're adding L2 support

Foivos (approved):

milapsheth (approved):

PR #284, opened 2024-07-22
feat: add rpc test to match logIndex

blockchainguyy:
https://axelarnetwork.atlassian.net/browse/AXE-4496

### DIFF
--- a/contracts/test/TestRpcCompatibility.sol
+++ b/contracts/test/TestRpcCompatibility.sol
@@ -7,15 +7,6 @@ contract TestRpcCompatibility {
 
     event ValueUpdated(uint256 indexed value);
     event ValueUpdatedForSubscribe(uint256 indexed value);
-    event ContractCallWithToken(
-        address indexed sender,
-        string destinationChain,
-        string destinationContractAddress,
-        bytes32 indexed payloadHash,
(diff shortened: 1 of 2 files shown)


contracts/test/TestRpcCompatibility.sol
```diff
@@ -21,4 +30,15 @@ contract TestRpcCompatibility {
         subscribeValue = newValue;
         emit ValueUpdatedForSubscribe(newValue);
     }
+
+    function emitCallContractWithToken(
```
milapsheth, line 34: Why add another event? We already have a method that emits an event, and we don't care about the event content for this test

test/RpcCompatibility.js
```diff
@@ -384,6 +384,33 @@ describe('RpcCompatibility', () => {
         });
     });
 
+    it('should match the fetched logIndex value', async function () {
```
milapsheth, line 387:
```suggestion
    it('should return consistent logIndex values between eth_getLogs and eth_getTransactionReceipt', async () => {
```
need to make the purpose for this specific test clear, and use `async () =>` for consistency

milapsheth (approved):

milapsheth (approved):

PR #287, opened 2024-08-23
chore(test): add case for validating parent hash

sdaveas:
Add tests to validate block parent hashes. In the light of Flow behaviour explained as per [this](https://interoplabs.slack.com/archives/C021B409SG5/p1724405403164199?thread_ts=1724405178.231209&cid=C021B409SG5) thread.

### DIFF
--- a/test/RpcCompatibility.js
+++ b/test/RpcCompatibility.js
@@ -133,6 +133,7 @@ describe('RpcCompatibility', () => {
                     finalizedBlockNumber = await provider.send('eth_getBlockByNumber', ['finalized', false]);
                 } catch (error) {
                     console.error('Failed to retrieve finalized block number:', error);
+                    throw error;
                 }
             });
 
--- a/test/RpcCompatibility.js
+++ b/test/RpcCompatibility.js
@@ -263,7 +263,7 @@ describe('RpcCompatibility', () => {
         });
 
         it('should have valid parent hashes', async () => {
-            validParentHashes('latest')
+            validParentHashes('latest');
         });
     });
 
--- a/test/RpcCompatibility.js
+++ b/test/RpcCompatibility.js
@@ -264,6 +264,7 @@ describe('RpcCompatibility', () => {
         });
 
         it('should have valid parent hashes', async () => {
+            // Note: If chain doesn't have instant finality, reorgs could cause this to fail
             validParentHashes('latest');
         });
     });
--- a/test/RpcCompatibility.js
+++ b/test/RpcCompatibility.js
@@ -39,9 +39,9 @@ describe('RpcCompatibility', () => {
         expect(timeDifference).to.be.lessThanOrEqual(maxDifference);
     }
 
-    async function validParentHashes(tag) {
+    async function validParentHashes(blockTag) {
         const withTransaction = false;
-        const block = await provider.send('eth_getBlockByNumber', ['latest', withTransaction]);
+        const block = await provider.send('eth_getBlockByNumber', [blockTag, withTransaction]);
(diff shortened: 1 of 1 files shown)


test/RpcCompatibility.js
```diff
@@ -243,6 +243,19 @@ describe('RpcCompatibility', () => {
...
             checkBlockTimeStamp(parseInt(block.timestamp, 16), 12000);
         });
+
+        it('blocks have valid parent hashes', async () => {
+            const withTransaction = false;
+            const block = await provider.send('eth_getBlockByNumber', ['latest', withTransaction]);
+            const parentBlock = await provider.send('eth_getBlockByHash', [block.parentHash, withTransaction]);
+
+            console.log('Block:', block);
+            console.log('Parent block:', parentBlock);
```
milapsheth, line 253: No need to log unless debugging

test/RpcCompatibility.js
```diff
@@ -243,6 +243,19 @@ describe('RpcCompatibility', () => {
             checkBlock(block, false);
             checkBlockTimeStamp(parseInt(block.timestamp, 16), 12000);
         });
+
+        it('blocks have valid parent hashes', async () => {
```
milapsheth, line 247: Add another test for finalized block (if the chain supports it, that one should work too)

milapsheth (approved):

test/RpcCompatibility.js
sdaveas: oops forgot to cleanup thanks! addressed as part of 75eeb36acf34872bc010efaf682d245903bb553d

test/RpcCompatibility.js
sdaveas: sg 75eeb36acf34872bc010efaf682d245903bb553d

test/RpcCompatibility.js
```diff
@@ -39,6 +39,14 @@ describe('RpcCompatibility', () => {
         expect(timeDifference).to.be.lessThanOrEqual(maxDifference);
     }
 
+    async function validParentHashes(tag) {
+        const withTransaction = false;
+        const block = await provider.send('eth_getBlockByNumber', ['latest', withTransaction]);
```
milapsheth, line 44:
```suggestion
    async function validParentHashes(blockTag) {
        const withTransaction = false;
        const block = await provider.send('eth_getBlockByNumber', [blockTag, withTransaction]);
```

test/RpcCompatibility.js
```diff
@@ -117,11 +125,19 @@ describe('RpcCompatibility', () => {
...
+        describe('supports finalized tag', () => {
+            let finalizedBlockNumber;
 
-            try {
-                const finalizedBlockNumber = await provider.send('eth_getBlockByNumber', ['finalized', false]);
+            before(async () => {
+                try {
+                    finalizedBlockNumber = await provider.send('eth_getBlockByNumber', ['finalized', false]);
+                } catch (error) {
+                    console.error('Failed to retrieve finalized block number:', error);
```
milapsheth, line 135:
```suggestion
                    console.error('Failed to retrieve finalized block number:', error);
                    throw error;
```
throw again to fail, no point in running the below tests if it's not supported

test/RpcCompatibility.js
```diff
@@ -243,6 +261,10 @@ describe('RpcCompatibility', () => {
             checkBlock(block, false);
             checkBlockTimeStamp(parseInt(block.timestamp, 16), 12000);
         });
+
+        it('should have valid parent hashes', async () => {
+            validParentHashes('latest');
```
milapsheth, line 266:
```suggestion
            // Note: If chain doesn't have instant finality, reorgs could cause this to fail
            validParentHashes('latest');
```

PR #296, opened 2025-04-04
test: manually track balances to fix multiple transactions error

AttissNgo:
## Issue
Tests were failing with the error "Multiple transactions found in block" when running against the Plume testnet. This happens because Hardhat's `.changeEtherBalance()` matcher expects only one transaction per block to accurately track balance changes, but on live networks multiple transactions can be included in the same block.

## Changes Made
- Replaced all instances of `.changeEtherBalance()` matcher with manual balance tracking
- Added explicit balance checks before and after transactions
- Added debugging note to ReadMe about `selfdestruct` behavior post-Cancun

### DIFF
--- a/test/gmp/AxelarGasService.js
+++ b/test/gmp/AxelarGasService.js
@@ -117,7 +117,6 @@ describe('AxelarGasService', () => {
                 .and.to.emit(testToken, 'Transfer')
                 .withArgs(userWallet.address, gasService.address, gasFeeAmount);
 
-            ////
             const balanceBefore1 = await ethers.provider.getBalance(gasService.address);
 
             await expect(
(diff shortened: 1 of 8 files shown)


h4writer:
This is ready to merge.
- I have fixed the cursor bugbot remarks.
- Fixed an additional `.changeEtherBalance()` which was forgotten.
- Made sure the code correctly works post-cancun.
- Plume doesn't fail on `Error: Multiple transactions found in block` anymore.
- Ran `npm run lint` successfully. No idea why it fails to run here.

_Plume still fails on other exceptions. 
E.g. `AssertionError: Expected event "TokenDeployed" to be emitted, but it wasn't`
Does Plume not have a full deployment of axelar? Investigation is not part of this PR_

test/gmp/AxelarGasService.js
```diff
@@ -152,8 +158,10 @@ describe('AxelarGasService', () => {
                     amount,
                     nativeGasFeeAmount,
                     userWallet.address,
-                )
-                .and.to.changeEtherBalance(gasService, nativeGasFeeAmount);
+                );
+
+            const balanceAfter2 = await ethers.provider.getBalance(gasService.address);
+            expect(balanceAfter2.sub(balanceBefore2)).to.equal(nativeGasFeeAmount);
```
jakovmitrovski: Not really an issue, but Im wondering if it's more intutivie to 
`expect(balanceBefore2.add(nativeGasFeeAmount)).to.equal(balanceAfter2);` . Wdyt?

PR #297, opened 2025-04-24
feat: add future block compatibility test

blockchainguyy:
This PR adds hardhat test to query future block in RPC compatibility test.

JIRA: https://axelarnetwork.atlassian.net/browse/AXE-9145

### DIFF
--- a/test/RpcCompatibility.js
+++ b/test/RpcCompatibility.js
@@ -3,7 +3,7 @@
 const chai = require('chai');
 const { ethers, network } = require('hardhat');
 const {
-    utils: { hexValue, getAddress, keccak256, id },
+    utils: { hexValue, getAddress, keccak256, id, hexlify, randomBytes },
     Wallet,
     BigNumber,
 } = ethers;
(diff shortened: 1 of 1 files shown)


test/RpcCompatibility.js
```diff
@@ -484,4 +484,21 @@ describe('RpcCompatibility', () => {
             await checkReceipt(receipt, newValue);
         });
     });
+
+    it('should throw error when querying eth_getLogs with a future blockHash', async () => {
+        const currentBlockNumber = await provider.getBlockNumber();
+        const futureBlockNumber = currentBlockNumber + 1000;
+
+        const fakeFutureBlockHash = ethers.utils.hexZeroPad(ethers.utils.hexlify(futureBlockNumber), 32);
```
milapsheth, line 492: There's no way to predict a future block hash so this is effectively checking for an unknown block hash and not a future one? Also, can random 32 bytes suffice instead of trying to encode a future block number?

test/RpcCompatibility.js
```diff
@@ -484,4 +484,21 @@ describe('RpcCompatibility', () => {
             await checkReceipt(receipt, newValue);
         });
     });
+
+    it('should throw error when querying eth_getLogs with a future blockHash', async () => {
```
milapsheth, line 488: What about querying a future block number? Should we add coverage for that?

test/RpcCompatibility.js
```diff
@@ -484,4 +484,21 @@ describe('RpcCompatibility', () => {
...
+
+        const fakeFutureBlockHash = ethers.utils.hexZeroPad(ethers.utils.hexlify(futureBlockNumber), 32);
+
+        const params = [
+            {
+                blockHash: fakeFutureBlockHash,
+            },
+        ];
+
+        if (!isHardhat) {
```
milapsheth, line 500: What does hardhat do? Can its behavior be tuned by the hardhat config to match actual chains?

PR #320, opened 2026-06-18
fix: revert native transfers to bricked gateway with Bricked

haoshengzhen:
## Summary

- Add a `receive()` function to `BrickedAxelarGateway` that reverts with `Bricked()`.
- Tighten the native value transfer test to assert the same custom error used by the fallback path.

## Motivation

`BrickedAxelarGateway` is intended to make the gateway permanently unusable after upgrade, with calls reverting through the `Bricked()` custom error. Calls with calldata already hit the fallback and return `Bricked()`, but native value transfers with empty calldata reverted without a reason because there was no payable `receive()` path.

This keeps the bricked behavior consistent for plain native transfers as well.

## Tests

- `npx hardhat test test/BrickedAxelarGateway.js`
- `npm run lint`
- `npm run build`
- `git diff --check`

### DIFF
--- a/contracts/BrickedAxelarGateway.sol
+++ b/contracts/BrickedAxelarGateway.sol
@@ -41,6 +41,10 @@ contract BrickedAxelarGateway {
         bytes calldata /* setupParams */
     ) external pure {}
 
+    receive() external payable {
+        revert Bricked();
+    }
+
     // solhint-disable-next-line payable-fallback
     fallback() external {
         revert Bricked();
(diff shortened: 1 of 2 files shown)


haoshengzhen:
@MakisChristou @jakovmitrovski @milapsheth Hi, Could you please review this PR at your convenience? Thank you very much.

jakovmitrovski:
Hi! Thanks for putting this together! 

I think we'll leave it as-is for now though. The behavior is effectively the same either way. The contract already rejects native transfers. For a contract that's meant to be permanently dead, we'd rather not add a payable function, and it also re-trips Slither's `locked-ether` detector (which is actually why we dropped this same `receive()` back in #318). 

We really appreciate your contribution!

PR #324, opened 2026-08-27
refactor: correct implementation variable spelling

wangjingshuiku:
Correct two spelling errors in the Solidity sources:

- rename the fallback's local `implementaion_` variable to `implementation_` and update its assembly reference;
- correct “registed” to “registered” in `AxelarGateway` NatSpec.

The local rename does not affect storage, ABI, bytecode control flow, or the address passed to `delegatecall`.

Validation:

    `npx prettier --check contracts/util/Proxy.sol contracts/AxelarGateway.sol`


wangjingshuiku:
@MakisChristou Could you please take a look when you have a chance? Thanks!

